Issue #7, opened 2026-05-18
[RFC] Clean SGLang Runtime Code from vime


gcanlin:
# RFC: Clean SGLang Runtime Code from vime

## Summary

vime was originally forked from slime, whose rollout path historically used SGLang as the default inference backend. Now that the vLLM adaptation has landed, vime's runtime target should converge on vLLM and stop maintaining SGLang compatibility.

This RFC proposes cleaning the SGLang runtime code from vime's slime-derived codebase. The cleanup rule is not "delete anything that mentions SGLang." The correct order is:

1. Prefer an equivalent capability in vLLM, vllm-router, or vLLM RLHF weight transfer.
2. If the dependency is generic infrastructure rather than SGLang semantics, reimplement it as a vime-owned utility.
3. Delete only code that is truly SGLang-only: SGLang parameters, protocols, server lifecycle, external SGLang compatibility, and legacy debug entry points.
4. Keep SGLang Docker image build and patch assets, but mark them as docker-only / legacy build assets. They must not imply that vime runtime supports SGLang.

The target state is: vime runtime no longer imports `sglang`, `sglang_router`, or `sglang.srt.*`; the default rollout path is vLLM-only; SGLang is no longer a user-selectable backend.

## Goals

1. Remove the SGLang runtime compatibility surface: `--rollout-backend=sglang`, SGLang engine actors, SGLang rollout functions, and external SGLang server/router support.
2. Migrate valuable capabilities to vLLM equivalents: engine lifecycle, router integration, rollout HTTP APIs, sleep/wake, prefix cache reset, weight update, routed experts replay, and related flows.
3. Move generic utilities into vime-owned code: tensor buckets, multiprocessing serializers, torch reduction patches, quantization helpers, HTTP/router helpers, and deployment config dataclasses.
4. Remove the `--sglang-*` argument system. Preserved capabilities should be renamed to `--vllm-*` or engine-agnostic `--rollout-*` / `--router-*` options.
5. Remove SGLang-specific tests, examples, CI suites, and current user-facing docs. Valuable coverage should be migrated to vLLM tests.
6. Keep SGLang build assets in the Dockerfile design, while clearly separating them from runtime support.

## Non-Goals

1. Do not keep SGLang as a backend, provide fallback behavior, or add old-argument shims.
[...]

PR #9, opened 2026-05-18
[dockerfile]: swich to vllm==0.21.0

princepride:
## Purpose

Removed all SGLang-related Docker dependencies and replaced with vllm==0.21.0.

@gcanlin @aoshen02 I don't know how to test it because right now **vime** is still heavily dependent on SGLang.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,122 +1,87 @@
-ARG SGLANG_IMAGE_TAG=v0.5.10.post1
-FROM slimerl/sglang:${SGLANG_IMAGE_TAG} AS sglang
+ARG BASE_IMAGE=vllm/vllm-openai:v0.21.0
+FROM ${BASE_IMAGE}
 
-# ======================================== Arguments =============================================
+# vLLM is the inference/runtime dependency source of truth for this image.
+# SGLang is intentionally not installed in this build path.
+ARG VLLM_VERSION=0.21.0
+ARG MEGATRON_COMMIT=3714d81d418c9f1bca4594fc35f9e8289f652862
+ARG PATCH_VERSION=v0.5.9
+ARG INSTALL_VLLM=0
+ARG VLLM_PIP_INSTALL_ARGS=
 
-ARG PATCH_VERSION=latest
-ARG MEGATRON_COMMIT=1dcf0dafa884ad52ffb243625717a3471643e087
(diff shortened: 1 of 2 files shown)


gcanlin:
> ## Purpose
> Removed all SGLang-related Docker dependencies and replaced with vllm==0.21.0.
> 
> @gcanlin @aoshen02 I don't know how to test it because right now **vime** is still heavily dependent on SGLang.

I guess we need to wait https://github.com/vllm-project/vime/issues/7 if we want to remove the dependencies from SGLang. But we can add vllm installation first. And after https://github.com/vllm-project/vime/issues/7 done, we can remove SGLang totally in Dockerfile.

aoshen02:
Is there any cuda/pytorch/transformers version conflict between original configuration and what vLLM 0.21.0 needs?

aoshen02:
To let training not break, I think should still install sglang with --no-deps for now.

docker/Dockerfile
```diff
@@ -1,122 +1,87 @@
...
-    cp flash_attn_interface.py $python_path/flash_attn_3/flash_attn_interface.py && \
-    rm -rf flash-attention/
-
-RUN pip install git+https://github.com/ISEEKYAN/mbridge.git@89eb10887887bc74853f89a4de258c0702932a1c --no-deps
-
-RUN pip install flash-linear-attention==0.4.1
-RUN pip install tilelang -f https://tile-ai.github.io/whl/nightly/cu128/
-
-# TE does not have wheel on cuda 13 yet, thus need to install from source
-RUN if [ "${ENABLE_CUDA_13}" = "1" ]; then \
```
aoshen02, line 41: We need TE, it's a dependency of mcore

princepride:
> Is there any cuda/pytorch/transformers version conflict between original configuration and what vLLM 0.21.0 needs?

So we have better fallback to 0.20.2 ?

aoshen02:
https://github.com/vllm-project/vime/pull/12 I 

> > Is there any cuda/pytorch/transformers version conflict between original configuration and what vLLM 0.21.0 needs?
> 
> So we have better fallback to 0.20.2 ?

Not sure.

aoshen02:
https://github.com/vllm-project/vime/pull/12 I have a follow up PR that fix some problems, please check.

PR #10, opened 2026-05-18
refactor(vllm): consolidate scattered vllm args via wholesale AsyncEngineArgs import

aoshen02:
## Summary

Mirror the sglang_utils/arguments.py pattern for vllm — introduce a single source of truth for all `--vllm-*` flags and let `launch_server_process` auto-forward them to the `vllm serve` subprocess. Drop the dead `--rollout-backend` flag (upstream slime has no equivalent; vime is vllm-only). All router CLI knobs sit under a unified `--router-*` namespace.

### New file
- **`slime/backends/vllm_utils/arguments.py`** — wholesale wrapper that monkey-patches `parser.add_argument` AND `parser.add_argument_group` (vllm uses arg groups internally) so `AsyncEngineArgs.add_cli_args(parser)` auto-registers ~190 vllm CLI flags with `--vllm-` prefix and `vllm_` dest. Adds vime orchestration extras (`--router-ip/port/request-timeout-secs`, `--vllm-server-concurrency`) plus the `--vllm-weight-sync-packed` mutex group. Mirrors sglang's `add_xxx_arguments` / `validate_args` / `xxx_parse_args` triple. `validate_args` mirrors slime's pattern: aliases `args.vllm_dp_size` / `args.vllm_pp_size`, derives `args.vllm_tp_size = rollout_num_gpus_per_engine // vllm_pp_size`. `vllm_parse_args` mirrors slime's `sglang_parse_args` and computes a `vllm_tensor_parallel_size` default from a temp parser.

### `slime/utils/arguments.py`
- Delete scattered `--vllm-gpu-memory-utilization`, `--vllm-enforce-eager`, `--vllm-weight-sync-mode`, `--vllm-weight-sync-packed` (all moved into `vllm_utils/arguments.py`)
- Delete `--rollout-backend` flag entirely (upstream slime has no such flag; vime vllm-only)
- Drop top-level imports of `sglang_router.launch_router.RouterArgs` and `slime.backends.sglang_utils.arguments` (vime is vllm-only after this PR; sglang code paths are unreachable from the main parser)
- `add_router_arguments` now registers `vllm_router.RouterArgs.add_cli_args(parser, use_router_prefix=True, exclude_host_port=True)` — surfaces the full vllm-router CLI (`--router-policy`, `--router-cache-threshold`, ~30 more) on the main parser
- Add Phase 1 vllm pre-parse in `parse_args()` (sglang Phase 1a is removed)
- Delete the dead `--use-slime-router` flag and its deprecated-warning handler in `slime_validate_args`
- YAML config (`--custom-config-path`) now updates `args._vllm_user_provided` so YAML overrides defeat vime-preferred defaults

### `slime/backends/vllm_utils/vllm_engine.py`
- `launch_server_process` now uses `get_vllm_cli_action_table` to auto-forward `args.vllm_*` (no more hardcoded `cmd += ["--gpu-memory-utilization", ...]` per flag)
- Primitive vs structured detection: dataclass values like `WeightTransferConfig` are JSON-serialized via a robust encoder handling sets/enums/Path; scalar `nargs='+'/'*'` YAML values are normalized to single-element lists
- Raw-CLI-string capture (`args._vllm_raw_values`): dataclass-backed flags like `--vllm-compilation-config` round-trip via the user's original JSON instead of `asdict()` snapshots that vllm's subprocess parser would reject
- Reads `args.router_ip / router_port` (renamed; see "Router CLI unification" below)

### `slime/backends/megatron_utils/update_weight/update_weight_from_distributed.py` and `update_weight_from_tensor.py`
- Remove dead `use_vllm` parameter from `update_weights_from_distributed` (PR #10 hardcodes VLLMEngine, so `use_vllm=True` at every callsite). The sglang `else: dist.broadcast(...)` fallback branch is removed. `group` is always a `_NcclBridge` now; `packed` selects between `send_weights_packed` and `broadcast_tensors`.

### Router CLI unification (`--router-*`)

**Renamed flags** (and the corresponding dests):

| before | after | reads |
|---|---|---|
| `--vllm-router-ip` | `--router-ip` | `args.router_ip` |
| `--vllm-router-port` | `--router-port` | `args.router_port` |
| `--vllm-router-request-timeout-secs` | `--router-request-timeout-secs` | `args.router_request_timeout_secs` |

All readers updated (`vllm_engine.py`, `ray/rollout.py`, `rollout/vllm_rollout.py`).

**Why drop the `--vllm-` prefix on these three flags:**

Before this PR, `train.py --help` showed two unrelated-looking groups of router flags:

- `--vllm-router-ip`, `--vllm-router-port`, `--vllm-router-request-timeout-secs` — registered manually by vime's `add_vllm_router_arguments`
- `--router-policy`, `--router-cache-threshold`, `--router-api-key`, ~30 more — registered by `vllm_router.RouterArgs.add_cli_args(use_router_prefix=True)`

The two prefixes were a leak of *implementation* (vime registers ip/port/timeout; the vllm_router package registers everything else). From a user's perspective, all of these are just "router knobs" and belong in one namespace. slime's mixed prefix (`--sglang-router-ip` + `--router-policy`) is historical accident — `sglang_router` happens to add its CLI with the `router-` prefix already, so slime had to pick a different prefix for the IP/port/timeout flags it adds on top.

PR #10 already removes `--rollout-backend` and hardcodes `VLLMEngine`, so vime has a single backend. Without backend disambiguation to justify it, the `--vllm-` prefix on these three orchestration flags only preserves an internal boundary users shouldn't need to understand. Consistency within vime is worth more than mirroring slime's historical split.

### Drop sglang back-compat migration
- Delete `_migrate_legacy_sglang_flags` and `_VIME_LEGACY_VLLM_DEFAULTS` from `vllm_utils/arguments.py` (closed Gemini bot's silent-skip bug and CalvinXKY's "sglang args still pre-parsed" comment in one move)
- `--sglang-router-*` flags are no longer parsed or migrated

## Python 3.12 argparse compat

`AsyncEngineArgs.add_cli_args` passes `deprecated` / `deprecated_aliases` kwargs to `add_argument`. argparse on Python <3.13 with `BooleanOptionalAction` rejects them. Wrapper now strips these kwargs before forwarding to the underlying `add_argument`. (CalvinXKY's issue-level bug report.)

## Misc

- `get_vllm_cli_action_table()` now caches its result at module level (CalvinXKY's perf note: it was being rebuilt twice in `launch_server_process`).

## Why these particular pieces (cache management, structured-value pruning, raw-value capture)

All justified by architectural differences: vllm runs as a subprocess (sglang in-process), uses argument groups internally, and has a vime-preferred GPU memory default that diverges from vllm's library default.

## Validation

```
python3 -c "
from slime.backends.vllm_utils.arguments import vllm_parse_args, get_vllm_cli_action_table
ns = vllm_parse_args()
print('# vllm_* attrs:', len([k for k in vars(ns) if k.startswith('vllm_')]))
print('# forwardable:', len(get_vllm_cli_action_table()))
print('router_ip default:', getattr(ns, 'router_ip', 'MISSING'))
print('router_port default:', getattr(ns, 'router_port', 'MISSING'))
print('gpu_mem default:', ns.vllm_gpu_memory_utilization)  # 0.92 (vllm-side; vime applies 0.55 in launch_server_process)
"
```

`pytest -m unit -v tests/test_chunked_gae.py tests/test_megatron_argument_validation.py tests/test_qwen3_5_mtp_bridge_mapping.py tests/plugin_contracts/` — **15/15 PASS** on this branch. (`tests/test_qwen3_linear_attention_cu_seqlens.py` is excluded; its 2 failures are caused by a missing `flash-linear-attention` install in the test image and are unrelated to this PR.)

## Test plan

- [x] `python -c "from slime.utils.arguments import parse_args"` imports cleanly
- [x] `pytest -m unit` 15/15 PASS in the targeted suite
- [ ] End-to-end Qwen2.5-0.5B training: loss decreases, weight sync works, no router crash
- [ ] User-supplied vllm flag (e.g. `--vllm-max-model-len 8192`) reaches `vllm serve` subprocess

## Breaking changes

- `--rollout-backend` removed (vllm-only)
- `--sglang-router-*` no longer recognized; use `--router-*` equivalents
- `--vllm-router-ip / --vllm-router-port / --vllm-router-request-timeout-secs` renamed to `--router-ip / --router-port / --router-request-timeout-secs`
- `--use-slime-router` removed
- Launch scripts that read `args.vllm_router_*` or `args.sglang_router_*` must update to `args.router_*`

### DIFF
--- a/slime/backends/vllm_utils/arguments.py
+++ b/slime/backends/vllm_utils/arguments.py
@@ -22,6 +22,18 @@ from slime.utils.http_utils import _wrap_ipv6
 logger = logging.getLogger(__name__)
 
 
+# argparse 3.13+ accepts ``deprecated`` / ``deprecated_aliases`` on add_argument;
+# vLLM may pass them while using BooleanOptionalAction on Python 3.12, which rejects them.
+# Strip these kwargs on older Python so AsyncEngineArgs.add_cli_args() doesn't crash.
+_ARGPARSE_UNSUPPORTED_KWARGS = frozenset({"deprecated", "deprecated_aliases"})
+
+
+def _strip_unsupported_argparse_kwargs(kwargs: dict) -> dict:
+    if sys.version_info >= (3, 13):
+        return kwargs
+    return {k: v for k, v in kwargs.items() if k not in _ARGPARSE_UNSUPPORTED_KWARGS}
+
+
 def _detect_user_provided_dests(parser, argv: list[str]) -> tuple[set[str], dict[str, str]]:
     """Return (user_provided, raw_values) extracted from ``argv``.
 
@@ -82,10 +94,11 @@ SKIPPED_DESTS = [
     "seed",
     # dtype: vime --fp16 / training config owns this
     "dtype",
-    # distributed topology: rollout_num_gpus_per_engine + actor layout own this
+    # tp_size is fully owned by the orchestrator (rollout_num_gpus_per_engine
+    # // pp_size) — see validate_args. pipeline_parallel_size and
+    # data_parallel_size remain user-controllable and auto-forward to the vllm
+    # subprocess when set; mirrors slime sglang_utils which exposes both via CLI.
     "tensor_parallel_size",
-    "pipeline_parallel_size",
-    "data_parallel_size",
     # network: engine launcher decides per-engine port/host
     "port",
     "host",
@@ -95,24 +108,30 @@ SKIPPED_DESTS = [
 
 
 def add_vllm_router_arguments(parser):
-    """vime's vllm-router orchestration flags (not in AsyncEngineArgs)."""
+    """vime's router orchestration flags (where to reach the router; not in vllm-router's CLI).
+
+    Named without a ``--vllm-`` backend prefix so they sit alongside
+    ``vllm_router.RouterArgs`` knobs (``--router-policy``, ``--router-cache-threshold``,
+    …) under a single ``--router-*`` namespace. PR #10 hardcodes a single vllm backend,
+    so backend-disambiguation prefix would only leak implementation.
+    """
     parser.add_argument(
-        "--vllm-router-ip",
+        "--router-ip",
         type=str,
         default=None,
-        help="IP address of the vLLM router.",
+        help="IP address of the router (where vime connects to send rollout requests).",
     )
     parser.add_argument(
-        "--vllm-router-port",
+        "--router-port",
         type=int,
         default=None,
-        help="Port of the vLLM router.",
+        help="Port of the router.",
     )
     parser.add_argument(
-        "--vllm-router-request-timeout-secs",
+        "--router-request-timeout-secs",
         type=int,
         default=14400,
-        help="Timeout for requests to the vLLM router in seconds.",
+        help="Timeout (seconds) for HTTP requests vime makes to the router.",
     )
     return parser
 
@@ -151,6 +170,9 @@ def _make_add_argument_wrapper(target_add_argument):
         if "dest" in new_kwargs and isinstance(new_kwargs["dest"], str):
             if not new_kwargs["dest"].startswith("vllm_"):
                 new_kwargs["dest"] = f"vllm_{new_kwargs['dest']}"
+
+        # Strip argparse kwargs that Python <3.13 rejects (vLLM passes them on newer Python).
+        new_kwargs = _strip_unsupported_argparse_kwargs(new_kwargs)
 
         return target_add_argument(*new_flags, **new_kwargs)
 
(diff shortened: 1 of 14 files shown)


slime/ray/rollout.py
```diff
@@ -954,70 +949,56 @@ def addr():
...
 
-    use_vllm_router = getattr(args, "rollout_backend", "vllm") == "vllm"
     from slime.utils.http_utils import run_router
 
-    if use_vllm_router:
-        router_args = _vllm_router_args_from_cli(args)
-        impl = "vllm"
-    else:
-        from sglang_router.launch_router import RouterArgs
-
```
CalvinXKY, line 977: Main parser still registers sglang RouterArgs (--sglang-router-*), while rollout reads args.vllm_router_* and builds vllm RouterArgs.from_cli_args. Migration covers IP/port, but other router knobs may still only exist on sglang_router_*. Consider registering vllm-router flags on the main parser (or document that only migrated fields work).

slime/utils/arguments.py  L9, L1018–1025 RouterArgs.add_cli_args（sglang_router）

slime/ray/rollout.py
```diff
@@ -113,14 +113,9 @@ def start_engines(self, port_cursors: dict[int, int] | None = None) -> tuple[lis
 
         pg, reordered_bundle_indices, reordered_gpu_ids = self.pg
 
-        if getattr(self.args, "rollout_backend", "sglang") == "vllm":
-            from slime.backends.vllm_utils.vllm_engine import VLLMEngine
+        from slime.backends.vllm_utils.vllm_engine import VLLMEngine
 
-            RolloutRayActor = ray.remote(VLLMEngine)
-        else:
-            from slime.backends.sglang_utils.sglang_engine import SGLangEngine
```
CalvinXKY, line 121: This PR hardcodes VLLMEngine but still always pre-parses and validates sglang args, and keeps sglang multi-model config paths. For a vllm-only fork that's confusing overhead; for a transitional repo, please clarify in the PR whether sglang code paths are dead and schedule cleanup.

slime/utils/arguments.py
CalvinXKY: slime_validate_args still forces --use-slime-router off in favor of sglang_router, but _start_router only launches vllm-router. Stale validation message; update or remove to avoid misleading operators.

slime/backends/vllm_utils/arguments.py
```diff
@@ -0,0 +1,361 @@
...
+        "vllm_router_ip",
+        "vllm_router_port",
+        "vllm_router_request_timeout_secs",
+        "vllm_server_concurrency",
+        "vllm_weight_sync_packed",
+    }
+)
+
+
+def get_vllm_cli_action_table():
```
CalvinXKY, line 329: get_vllm_cli_action_table() rebuilds the full parser via add_vllm_arguments() on every call. launch_server_process calls it twice. Consider caching the table at module level after first build.

CalvinXKY:
image

Bug: vllm_parse_args() crashes on Python 3.12 when vLLM passes deprecated= to add_argument

Location: slime/backends/vllm_utils/arguments.py, _make_add_argument_wrapper() (~L130–155), called from add_vllm_arguments() → AsyncEngineArgs.add_cli_args(parser) (~L216).

Repro: Python 3.12 + current vLLM; run python train.py (or vllm_parse_args()).

```
# argparse 3.13+ accepts ``deprecated`` / ``deprecated_aliases`` on add_argument;
# vLLM may pass them while using BooleanOptionalAction on Python 3.12, which rejects them.
_ARGPARSE_UNSUPPORTED_KWARGS = frozenset({"deprecated", "deprecated_aliases"})


def _strip_unsupported_argparse_kwargs(kwargs: dict) -> dict:
    if sys.version_info >= (3, 13):
        return kwargs
    return {k: v for k, v in kwargs.items() if k not in _ARGPARSE_UNSUPPORTED_KWARGS}
```

CalvinXKY:
Nit (non-blocking): rollout.py still uses SglangConfig / sglang_model_routers naming while engines are vLLM-only—consider a follow-up rename for clarity.

CalvinXKY:
The test file layout should be reorganized in a follow-up (or as part of a dedicated effort).

I suggest removing weight-transfer-related tests from this PR for now, and landing that coverage together with the reorganized test suite when the weight-sync changes are merged.

CalvinXKY:
Notice: some running arguments removed:
image

CalvinXKY:
Verified:

image

image

PR #16, opened 2026-05-19
Switch vllm_rollout endpoint /v1/completions to /inference/v1/generate

andakai:
## Description

This PR switches the vLLM text-only rollout path from the OpenAI-compatible /v1/completions endpoint to the SkyRL style token-only /inference/v1/generate endpoint.

The new path tokenizes prompts on the rollout client side, sends token_ids to vLLM, and parses generated token ids and per-token logprobs directly from the `/inference/v1/generate` response. This also aligns the vLLM rollout behavior with the existing SGLang text-only rollout style, where the rollout client owns tokenization and the server receives token ids.

More reasons about why use `inference/v1/generate` instead of `/v1/completions`:
- RL is sensitive to tokens. A word may be tokenized into different tokens. Better follow "Token in Token out" paradim. `/inference/v1/generate` is designed for this, see https://github.com/vllm-project/vllm/issues/22817 .
- It reduces dependency on OpenAI-compatible completion schema for RL training data construction.
- For multimodal rollout, /v1/completions is not a natural fit because it has no native multimodal request schema. See SkyRL's practice here: https://www.anyscale.com/blog/vision-language-model-reinforcement-learning-skyrl .
- Follow slime/sglang and SkyRL design.

## Changes

- Replace text-only vLLM rollout request URL from /v1/completions to /inference/v1/generate.
- Delete all related to `/v1/completions`.

**Note:**

This PR only targets the text-only vLLM rollout path. **Multimodal rollout is not part of this change.** The existing multimodal flow still uses: /v1/chat/completions/render -> /inference/v1/generate in SkyRL.

**Routed experts support is currently not supported in `/inference/v1/generate` in v0.20.2. So please don't merge this pr currently. It is in https://github.com/vllm-project/vllm/pull/39568.** Once it is released in next version. I will keep doing this pr.

### DIFF
--- a/slime/rollout/vllm_rollout.py
+++ b/slime/rollout/vllm_rollout.py
@@ -39,10 +39,10 @@ _PROCESSOR_PROMPT_KEYS = {"input_ids", "attention_mask"}
 
 
 def _coerce_flat_int_token_ids(ids: Any) -> list[int]:
-    """Turn tokenizer / processor output into a flat ``list[int]`` for vLLM ``/v1/completions`` JSON.
+    """Turn tokenizer / processor output into a flat ``list[int]`` for vLLM token-only JSON.
 
-    vLLM deserializes ``prompt`` as ``StringOrArray`` (Rust): a JSON string or a JSON array of integers.
-    Nested lists, numpy/torch scalars, or non-int elements cause ``422`` deserialization errors.
+    ``/inference/v1/generate`` requires ``token_ids: list[int]``. Nested lists, numpy/torch scalars,
+    or non-int elements cause ``422`` validation errors.
     """
     if ids is None:
         return []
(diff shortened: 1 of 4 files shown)


slime/rollout/vllm_rollout.py
```diff
@@ -408,6 +344,15 @@ def _build_inference_sampling_params(sampling_params: dict[str, Any]) -> dict[st
...
 
 
+def build_inference_payload(args: Namespace, sampling_params: dict[str, Any], token_ids: list[int]) -> dict[str, Any]:
+    """Build a vLLM ``/inference/v1/generate`` request body for token-only rollout."""
+    return {
+        "model": args.hf_checkpoint,
+        "token_ids": token_ids,
+        "sampling_params": _build_inference_sampling_params(sampling_params),
+    }
+
```
aoshen02: Do we need to have a wrapper on the function _build_inference_sampling_params?

aoshen02:
Please add some tests.

slime/rollout/vllm_rollout.py
andakai: This makes sense. Deleted this wrapper.

andakai:
Added a new test covered: Qwen3-0.6B, Qwen3-30B-A3B, Qwen3-30B-A3B-R3.

andakai:
Note that currently vllm v0.21.0 will fail at the `test_qwen3_30b_a3b_r3_vllm_inference_generate_endpoint` test.

PR #17, opened 2026-05-19
dockerfile: cu12 base for vllm rollout (vllm/vllm-openai:v0.21.0-cu129)

aoshen02:
## Summary

Switches vime's `docker/Dockerfile` base from `slimerl/sglang:v0.5.10.post1` (cu12 sglang custom image) to `vllm/vllm-openai:v0.21.0-cu129-ubuntu2404` (cu12 vllm official image). Same vllm version, same RLHF HTTP API, same router CLI — just an aligned base.

## Why

- vime's rollout backend is **vllm**, but main's Dockerfile inherits slime's `slimerl/sglang` base which already ships sglang + cu12 torch. Layering vllm on top is the wrong direction.
- The earlier vime cu13 image (PR #9 + PR #12 cumulative) used `vllm/vllm-openai:v0.21.0` default tag (cu130) and paid for it: 41.3 GB final size, 12-minute TE source compile against cu13 dev headers, plus an extra `apt install cuda-cudart-12-9` to make `torch_memory_saver`'s pypi wheel resolve (its hook is linked to libcudart.so.12).
- cu129 is the **same vllm 0.21.0** with everything natively on cu12. Final image **19.6 GB** (less than half), TE source build links cleanly, `torch_memory_saver` works as-shipped, no cuda-cudart-12 apt hack.

## What's installed beyond the cu13 vime image

- `flash-attn==2.7.4.post1` source-compiled (slime upstream parity)
- `flash-attn` 3 hopper kernels source-compiled (slime upstream parity)
- `apex` from source (slime upstream parity)

apex enables Megatron's fusion kernels — launchers no longer need `--no-rope-fusion --no-gradient-accumulation-fusion`.

## What stayed

- Megatron `1dcf0da` + slime megatron.patch (latest)
- mbridge / Megatron-Bridge / nvidia-modelopt
- flash-linear-attention 0.4.1 + tilelang (linear-attention models)
- `sglang` + `sglang-router` via `--no-deps` stub (vime's `slime/utils/arguments.py` still top-level imports `sglang_router`)
- IPython, numpy<2, nvidia-cudnn-cu12 9.16 pin
- pytest + python→python3 symlink
- `ENTRYPOINT []` reset (vllm/vllm-openai base sets `vllm serve` by default)
- requirements.txt filter to skip sglang lines

## Verification

Local build + Qwen3-0.6B 1-step vllm rollout train (mounted PR #10's code into the image, 4× H200):

| Stage | Result |
|---|---|
| Build (apt + TE + flash-attn 2/3 + apex + slime) | ✅ 24/24 layers, ~50 min, **19.6 GB** |
| `import vllm/sglang/apex/flash_attn/megatron.bridge` smoke | ✅ |
| Ray cluster + RolloutManager + 4× vllm engine | ✅ |
| Megatron actor load HF ckpt via `--megatron-to-hf-mode bridge` | ✅ |
| **Fusion kernels** (no `--no-*-fusion` flags) | ✅ |
| First weight update triggered | ✅ |
| Hits PR #10 documented PR #12 weight-sync gap (`start_weight_update must be called before update_weights`) | ✅ same boundary as the cu13 image — within scope of PR #10 / PR #12 split |

## Test plan

- [ ] CI builds the new Dockerfile end-to-end (uses `docker/patch/latest/megatron.patch`, no other repo deps changed)
- [ ] Smoke-import all key packages in the resulting image
- [ ] (Optional) Re-run an existing slime e2e test (e.g. `test_qwen3.5_0.8B_gsm8k_short.py`) under this image to confirm fusion kernel coverage
- [ ] PR #12's cu13 path can be retired once this lands, or kept as `ENABLE_CUDA_13=1` toggle if cu13 hardware is needed

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -117,8 +117,10 @@ RUN cd Megatron-LM && \
 
 # ====================================== Install main package ============================================
 
-COPY . /root/slime
-RUN cd /root/slime && \
+ARG VIME_COMMIT=main
+RUN git clone https://github.com/vllm-project/vime.git /root/slime && \
+    cd /root/slime && \
+    git checkout ${VIME_COMMIT} && \
     pip install -e . --no-deps
 
 RUN cd /root/slime/slime/backends/megatron_utils/kernels/int4_qat && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,122 +1,179 @@
-ARG SGLANG_IMAGE_TAG=v0.5.10.post1
-FROM slimerl/sglang:${SGLANG_IMAGE_TAG} AS sglang
+ARG BASE_IMAGE=vllm/vllm-openai:v0.21.0-cu129-ubuntu2404
+FROM ${BASE_IMAGE}
 
 # ======================================== Arguments =============================================
 
-ARG PATCH_VERSION=latest
+# MEGATRON_COMMIT + PATCH_VERSION track slime upstream's docker/Dockerfile
+# (slime 1dcf0da / patch latest as of 2026-02-14). vllm 0.21 cu129 base gives
+# us torch 2.11.0+cu129, libcudart.so.12 natively, vllm + vllm-router pre-installed.
 ARG MEGATRON_COMMIT=1dcf0dafa884ad52ffb243625717a3471643e087
+ARG PATCH_VERSION=latest
 
-ARG ENABLE_CUDA_13=0
+# Pinned to match slime upstream Dockerfile for parity (sglang here is install-on-top
+# via --no-deps; the cu129 base ships vllm, not sglang).
+ARG SGLANG_VERSION=0.5.10.post1
+ARG SLIME_TORCH_MEMORY_SAVER_COMMIT=d64a639
 
-# ======================================== Setup =============================================
+SHELL ["/bin/bash", "-euxo", "pipefail", "-c"]
 
 WORKDIR /root/
 
 # ======================================== Apt dependencies =============================================
 
-RUN apt update
-RUN apt install -y nvtop rsync dnsutils
+# TE 2.10 has no prebuilt wheel — always builds from source.  vllm/vllm-openai
+# base is an inference image, so the cu12 dev-header collection needed by the TE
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -1,10 +1,11 @@
-ARG SGLANG_IMAGE_TAG=v0.5.10.post1
-FROM slimerl/sglang:${SGLANG_IMAGE_TAG} AS sglang
+ARG BASE_IMAGE=vllm/vllm-openai:v0.21.0-cu129-ubuntu2404
```
CalvinXKY, line 1: The vLLM upgrade will pull in a newer PyTorch. Please confirm whether Megatron (and our Megatron-related deps) need a matching version bump, and verify that our existing patches still apply and run cleanly on the upgraded stack.

docker/Dockerfile
princepride, line 1: Do we also need update `docker/README.md`?

docker/Dockerfile
```diff
@@ -14,8 +15,16 @@ WORKDIR /root/
 
 # ======================================== Apt dependencies =============================================
 
+# vllm/vllm-openai base is an inference image — add cu12 dev headers + cmake/git
+# so TE / apex / flash-attn source builds find cusparse.h etc.
 RUN apt update
-RUN apt install -y nvtop rsync dnsutils
+RUN apt install -y nvtop rsync dnsutils git cmake \
```
princepride: Splitting this into two layers increases the image size. I think we can use `RUN apt-get update && apt-get install` instead.

docker/Dockerfile
```diff
@@ -96,27 +112,17 @@ RUN cd Megatron-LM && \
     fi && \
     rm megatron.patch
```
princepride: Although the original code patched megatron after an editable install😂, I still think we can change the order, because `megatron.patch` may modify any C++/CUDA files or `setup.py`

princepride (changes requested):
PTAL

princepride (approved):

knlnguyen1802:
@aoshen02 Should we also install vllm-router ?

docker/Dockerfile
aoshen02: fix

PR #38, opened 2026-05-26
docs: complete sglang→vllm rename across docs/ + README

aoshen02:
## Summary

Split from PR #18 (`gcl/clean-sglang`). One of 4 PRs splitting the original PR #18 by content area: **docs** / examples / tests+CI / core runtime. See [pr18-split-plan.md](pr18-split-plan.md) on PR #18.

**53 files. Pure documentation: no runtime code change.**

### What this PR does

- **`docs/en/` + `docs/zh/`** (49 files):
  - Rename `sglang-config.md` → `vllm-config.md` (1 each side)
  - Update all `--sglang-*` CLI flag examples to their `--vllm-*` equivalents
  - Rename references to SGLang as the rollout backend → vLLM
  - **Fix 3 broken upstream URLs** (each replacement verified to load):
    | Old (broken) | New (verified) |
    |---|---|
    | `blog.vllm.ai/2025/10/26/distributed-batch-invariance.html` | `vllm.ai/blog/2025-11-10-bitwise-consistent-train-inference` |
    | `docs.vllm.ai/.../features/spec_decode.html` (301 → 404) | `docs.vllm.ai/.../features/speculative_decoding/` |
    | `github.com/vllm-project/vllm-router` (404; repo does not exist) | `github.com/vllm-project/router` (the actual router repo) |

- **`docs/_static/image/`** (3 files): rename `sglang_config.png` → `vllm_config.png`; refresh `arch.png` to the vllm-router architecture.

- **`docs/{en,zh}/advanced/reproducibility.md`**: The en and zh files had their content **swapped** (en had Chinese body, zh had English body). Restored the correct mapping.

- **`docs/{en,zh}/blogs/introducing_slime.md`**: Remove the "Special thanks to the AMD GenAI - Foundation Model Team for Day-1 AMD hardware support" line. With AMD support removed in #36, the acknowledgement implied ongoing support that no longer exists.

- **`README.md` / `README_zh.md`**:
  - Expanded vLLM router argument prefix explanation (`--router-*` for vllm-router native flags vs `--vllm-router-*` for vime-side orchestration knobs)
  - Remove stale link to `docs/zh/advanced/slime_vllm_backend_design_v1.md` (deleted in commit `fe700db`)

### Independent of other splits

This PR can land at any time without coordination. See `pr18-split-plan.md` §"合并顺序无强约束".

## Test plan

[...]

PR #22, opened 2026-05-21
[feat] Support Colocated Weight Sync via CUDA IPC for  vime

knlnguyen1802:
co-author: @SamitHuang 

## Purpose

Enable colocated weight synchronization between the Megatron trainer and vLLM rollout engines running on the same GPU(s), using CUDA IPC (Inter-Process Communication) via Ray instead of NCCL distributed broadcast. This avoids cross-node traffic and reduces weight sync latency when trainer and inference engine share the same physical GPUs.

---

### New: `UpdateVLLMWeightFromTensor` (`slime/backends/megatron_utils/update_weight/update_weight_from_tensor_vllm.py`)

A new weight-update class following the [vLLM RLHF IPC approach](https://docs.vllm.ai/en/stable/examples/rl/rlhf_ipc/). It handles the full colocated sync lifecycle:

1. Megatron parameters are converted to HF format via `HfWeightIteratorBase`.
2. All trainer ranks call `IPCWeightTransferEngine.trainer_send_weights(send_mode="ray")`, creating a CUDA IPC handle per GPU.
3. Handles are all-gathered and merged so every vLLM worker can pick the handle matching its physical GPU UUID.
4. For non-colocated overflow engines, the existing NCCL distributed broadcast path (`update_weights_from_distributed`) is preserved unchanged.

The per-`update_weights` call lifecycle is:
```
release_memory_occupation(level=0)   # free KV cache + model weights
init_weight_transfer_engine          # first call only
start_weight_update
  [for each HF weight chunk]
    trainer_send_weights (IPC)
    update_weights_from_distributed  # overflow/distributed engines
finish_weight_update
resume_memory_occupation
```

### Modified: `slime/backends/vllm_utils/vllm_engine.py`

- **`update_weights(update_info)`** — new public Ray-callable entry point. Since `ipc_handles` are Python callables (closures from `monkey_patch_torch_reductions`) that cannot be JSON-serialized, they are serialized with `cloudpickle` and base64-encoded into `ipc_handles_pickled`, which the vLLM `IPCWeightTransferUpdateInfo` accepts when `VLLM_ALLOW_INSECURE_SERIALIZATION=1` is set.
- **`release_memory_occupation(level=1)`** — now accepts a `level` parameter. `level=0` releases both KV cache and model weights (required before IPC tensor injection); `level=1` (default, unchanged) releases KV cache only.
[...]

PR #40, opened 2026-05-26
tests + CI: complete sglang→vllm rename across tests/ and .github/

aoshen02:
## Summary

Split from PR #18 (`gcl/clean-sglang`). One of 4 PRs splitting the original PR #18 by content area: docs (#38) / examples (#39) / **tests+CI** / core runtime.

**42 files**.

### Why tests + CI bundled in one PR

The CI workflows in `.github/workflows/pr-test.yml` reference test file names by string in the matrix:

```yaml
matrix:
  info: [{"num_gpus": 8, "test_file": "test_qwen2.5_0.5B_sglang_config.py"}, ...]
```

PR #18 renames those tests to `_vllm_config.py`. If we split:
- Tests PR first → CI references old names that no longer exist → CI red
- CI PR first → CI references new names that don't exist yet → CI red

Bundled → no broken window.

#### (A) `tests/` (38 files)

- Mechanical CLI-flag rename: `--sglang-*` → `--vllm-*` equivalents (same table as PR #39).
- Variable rename: `SGLANG_ARGS` → `VLLM_ARGS` where present.
- **4 file renames** (all >85% similarity):
  | Old | New |
  |---|---|
  | `test_qwen2.5_0.5B_opd_sglang.py` | `test_qwen2.5_0.5B_opd_vllm.py` |
  | `test_qwen2.5_0.5B_sglang_config.py` | `test_qwen2.5_0.5B_vllm_config.py` |
  | `test_qwen2.5_0.5B_sglang_config_distributed.py` | `test_qwen2.5_0.5B_vllm_config_distributed.py` |
  | `test_sglang_config_mixed_offload.py` | `test_vllm_config_mixed_offload.py` |
  | `test_sglang_config_mixed_offload_ft.py` | `test_vllm_config_mixed_offload_ft.py` |
  | `tests/utils/test_sglang_config.py` | `tests/utils/test_vllm_config.py` |

- **2 new tests** for the IPC weight-transfer path (PR #22 territory):
  - `tests/test_update_weight_from_tensor.py`
  - `tests/unit/backends/megatron_utils/update_weight/test_update_weight_from_tensor.py`

#### (B) `.github/` (4 files)

- `workflows/conda-ci.yml`: `lmsysorg/sglang:v0.5.0rc0-cu126` → `inferactinc/public:vime-vllm-cu129-latest`.
- `workflows/pr-test.yml` + `pr-test.yml.j2`:
  - All container images: `slimerl/slime[-test]:latest` → vime image.
  - `e2e-test-sglang-config` job → `e2e-test-vllm-config` (label + matrix updated to renamed test files).
  - `e2e-test-megatron` + `e2e-test-image` matrices: `_opd_sglang.py` → `_opd_vllm.py`.
- `ISSUE_TEMPLATE/bug_report.yml`: drop `SGLang version (if relevant):` env field, add `vLLM version:` and `vllm-router version:` lines (`CUDA/ROCm` → `CUDA` change from #36 preserved).

### Sgl residue intentionally kept (4)

All are anti-regression assertions confirming sglang code paths are gone, not residual references:

- `tests/test_update_weight_from_tensor.py:753` — comment "The vLLM IPC implementation must NOT contain sglang-style Gloo gather code".
- `tests/unit/backends/vllm_utils/test_arguments.py:233-237` — 3 assertions that `--sglang-router-ip`, `--sglang-router-port`, and `sglang_router_ip` are NOT present in the argument parser.

### Test plan

- [ ] `git grep -iE 'sglang' .github/ tests/` returns only the 4 anti-regression assertion lines above
- [ ] CI matrix `test_file` strings match actual files on disk: `for f in $(grep -oE 'test_[a-zA-Z0-9._]+\.py' .github/workflows/pr-test.yml | sort -u); do test -f "tests/$f" || echo "MISSING: $f"; done`
- [ ] pre-commit passes (verified: black auto-formatted 8 files during prep, re-run is clean)

## Co-authors

Co-authored-by: Canlin Guo <canlinguosdu@gmail.com>

---

## Test execution status (h200-0 + h200-1, 16 × H200, 2026-05-26 → 05-27)

Each test was run inside `inferactinc/public:vime-vllm-cu129-latest` (PR #48 base + 2 local patches; see PR #48 description) with `--num-rollout 2` and `--use-wandb` stripped. r3 tests use a separately built `vime-vllm-r3:test` image (vllm nightly `cu129-nightly-ff712f6447093d07747c88680b9d006b119f5890`, which carries PR #39568).

Legend: ✅ pass · ❌ real bug · ⏭️ blocked on infra/setup · ⏸️ not run · 🔧 needs local patch to pass

### `tests/` (e2e)

- ✅ `test_qwen2.5_0.5B_short.py` — 4 GPU, basic smoke (PR #48 base + patches)
- ✅ `test_qwen2.5_0.5B_async_short.py` — 4 GPU, async rollout. 🔧 needs `os.environ.pop("http_proxy"/"HTTP_PROXY"/...)` → add `, None` default (4 calls); raises `KeyError` on hosts where the env var is unset. Trivial test fix.
- ✅ `test_qwen2.5_0.5B_ppo_critic_only_short.py` — 4 GPU
- ✅ `test_qwen3.5_0.8B_gsm8k_short.py` — 4 GPU
- ✅ `test_qwen3.5_0.8B_gsm8k_async_short.py` — 4 GPU. 🔧 same `os.environ.pop` fix as above.
- ✅ `test_qwen2.5_0.5B_debug_rollout_then_train.py` — 8 GPU, debug rollout + train two-phase
- ✅ `test_qwen3_0.6B_parallel_check.py` — 8 GPU. **Sweeps tp ∈ {1,2,4,8} × pp ∈ {1,2,4} × cp ∈ {1,2,4,8}** for `num_gpus ∈ {8,4,2}`. Required the 2 PR #48 follow-up patches below; ran the full 2-hour sweep without leader/UUID errors after the patches.
- ✅ `test_qwen3_0.6B_parallel_check.py` exposed two regressions in PR #48; both have fixes drafted:
  - leader-rank gate: `mpu.get_tensor_model_parallel_rank() == 0` → `rank == start` (slot leader). Megatron-TP=1 + rollout-TP>1 made every slot rank a coordinator → `start_weight_update called while a weight update is already active`.
  - IPC payload gather: `mpu.get_tensor_model_parallel_group()` → per-slot `dist.new_group(ranks=[start..end))`. When Megatron-TP != rollout-TP, the Megatron-TP group does not cover the engine slot; merged dict missed peer UUIDs → `IPC handle not found for GPU UUID ...`.
- ✅ `test_qwen3_4B_ppo.py` — 8 GPU PPO + critic colocate, IPC heavy
- ✅ `test_qwen3_4B_ppo_train_critic_only.py` — 8 GPU
- ✅ `test_qwen3_4B_ppo_disaggregate.py` — 8 GPU rollout-disaggregated. First attempt hit a Ray transient (`No available agent to submit job`); retry passed.
- ✅ `test_mimo_7B_mtp_only_grad.py` — 8 GPU, MTP gradient path
- ✅ `test_moonlight_16B_A3B.py` — 8 GPU MoE colocate, heavy IPC
- ✅ `test_quick_start_glm4_9B.py` — 8 GPU GLM smoke

- ❌ `test_qwen3_4B_ckpt.py` — 8 GPU. `ValueError: Cannot merge two lists with different lengths (81 and 52, encountered at level ('optimizer', 'param_state', 0, ...))` on the load step. Reproduces on **both** the old PR #18 branch and PR #48 base — not a PR #18/#48 regression. Looks like Megatron-LM `dist_checkpointing` cannot reload an optimizer state saved under a different TP/PP shape. Needs investigation in `slime/backends/megatron_utils/checkpoint.py` (does it serialize the shape on save?). **No fix PR yet** — orthogonal to the IPC RPC work.
- ❌ `test_qwen3_30B_A3B.py` — 8 GPU. `ValueError: The output_size of gate's and up's weight = 96 is not divisible by weight quantization block_n = 128`. `Qwen3-30B-A3B-FP8.moe_intermediate_size=768`; TP=8 gives 96 per rank. **Test config bug, not slime.** Use TP ∈ {1, 2, 3, 6} so `768/TP` stays divisible by 128. **No fix PR yet** (one-line `--tensor-model-parallel-size` change in the test).

- ⏭️ `test_glm4.7_30B_A3B_pd_mooncake.py` — needs Mooncake KV-store infra deployed on host + GLM-4.7-Flash download (HF repo name not confirmed). Also note: `--actor-num-nodes 1` despite the "dual node" name; this is single-node PD-disagg, not 2-node Ray.
- ⏭️ `test_qwen3.6_35B_A3B_pd_mooncake.py` — same Mooncake blocker. Model itself is on disk (`models-shared/Qwen3.6-35B-A3B`).
- ⏭️ `test_qwen3_30B_A3B_r3.py` — needs vllm-nightly image (built locally as `vime-vllm-r3:test`) but **also blocked by FP8 block_n config bug** above.
- 🔧 `test_moonlight_16B_A3B_r3.py` — 8 GPU. Tried with the new r3 image. Two import bugs surface on the r3 image:
  1. `slime/ray/rollout.py:16` — `from sglang.srt.constants import GPU_MEMORY_TYPE_*`. Easy patch: inline the 3 string constants (`"cuda_graph"/"kv_cache"/"weights"`).
  2. `slime/backends/megatron_utils/sglang.py:11,13` — `from sglang.srt.utils.patch_torch import monkey_patch_torch_reductions` (referenced via `sglang.monkey_patch_torch_reductions` in `update_weight_from_tensor.py`). Whole file plus its callers must be ported. Not blocking on the old image (sglang 0.5.10 is preinstalled there); only surfaces on the rebuilt nightly r3 image where sglang is absent. **No fix PR yet** — wider cleanup task.
- ⏸️ `test_gspo.sh` — never executed.

### `tests/` (unit / non-e2e, at the `tests/` root)

These 6 files live at `tests/` root but are pytest-style unit tests, not Ray-job e2e. Run as part of Phase 1 on the PR #18 branch.

- ✅ `test_chunked_gae.py`
- ✅ `test_megatron_argument_validation.py`
- ✅ `test_qwen3_5_mtp_bridge_mapping.py`
- ✅ `test_qwen3_linear_attention_cu_seqlens.py`
- ✅ `test_update_weight_from_tensor.py` — pytest harness around the IPC weight transfer; **distinct from** the same-named module under `slime/backends/megatron_utils/update_weight/`.
- ✅ `test_vllm_generate_endpoint.py`

### `tests/` files that are renamed-from-sglang by this PR (cannot run on base)

These don't exist on `origin/main`; they only show up in PR #48 base after PR #48 (which is built on this PR). One of them caught a real PR #18 NCCL bug when run on the older `vime-pr18-push` branch:

- ⏸️ `test_qwen2.5_0.5B_opd_vllm.py` — N/A on `origin/main`. Will run after this PR merges.
- ⏸️ `test_qwen2.5_0.5B_vllm_config.py` — N/A on `origin/main`. Will run after this PR merges.
- ❌ `test_qwen2.5_0.5B_vllm_config_distributed.py` — **Real bug on the PR #18 branch.** 4-GPU non-colocate distributed path: `torch.distributed.DistStoreError: Timed out, 3/4 clients joined`. The NCCL weight transfer rendezvous fails because only 3 of 4 expected clients connect. PR #48's IPC patches don't touch this path (it goes through `update_weights_from_distributed`, not `update_weights_from_tensor`). **No fix PR yet** — needs investigation in the distributed bridge connect/disconnect logic; likely a rank-count mismatch between trainer and rollout sides when not all rollout engines are colocated.
- ⏸️ `test_vllm_config_mixed_offload.py` — N/A on `origin/main`. Will run after this PR merges.
- ⏸️ `test_vllm_config_mixed_offload_ft.py` — N/A on `origin/main`. Will run after this PR merges.

### `tests/unit/`, `tests/utils/`, `tests/plugin_contracts/` (standalone)

Ran on PR #18 branch (HEAD `8e03368`) as Phase 1, 15 files: all pass except the 8 pre-existing failures in `tests/unit/backends/vllm_utils/test_vllm_engine.py` already documented in PR #48's commit message (`_weight_transfer_http_timeout`, `_response_json_or_fallback`, `server_host`, ...). Those 8 are unchanged by PR #48 and not in the IPC RPC scope — `main`: 8 failed / 24 passed; with PR #48: 8 failed / 26 passed (the 2 new IPC RPC unit tests).

### Failures requiring no fix PR

| Test | Reason | Fix kind |
|---|---|---|
| `test_qwen3_4B_ckpt.py` | Megatron `dist_checkpointing.merge` across TP shapes | upstream slime checkpoint logic |
| `test_qwen3_30B_A3B.py` | Test sets TP=8 incompatible with FP8 `block_n=128` and `moe_intermediate_size=768` | one-line `--tensor-model-parallel-size` in the test |

### Test bugs surfaced and fixes drafted

| File | Fix |
|---|---|
| `tests/test_qwen2.5_0.5B_async_short.py:117` and `tests/test_qwen3.5_0.8B_gsm8k_async_short.py` | `os.environ.pop("http_proxy")` → `os.environ.pop("http_proxy", None)` (× 4 variants: lower/upper `http_proxy`/`https_proxy`) |

### Open follow-up PRs against PR #48

The 2 IPC regressions caught by `test_qwen3_0.6B_parallel_check.py` need to land before PR #48 merges:

- `slime/backends/megatron_utils/update_weight/update_weight_from_tensor.py`:
  - `connect_rollout_engines`: build a `dist.new_group(ranks=range(start, end))` per engine slot **collectively across all trainer ranks**, store in `self._ipc_slot_group`; replace coordinator condition `mpu.get_tensor_model_parallel_rank() == 0` with `rank == start`.
  - `_send_hf_chunk_via_ipc`: replace `tp_group`/`tp_size` uses with the slot group/size.

### Test environment notes (not in this PR, applied at the harness layer)

- HF cache snapshot dirs were rewritten from symlinks to file-by-file symlinked real dirs to fix `SameFileError` in `hf` downloads (vllm's `_hf_hub_download_to_local_dir` calls `shutil.copyfile`).
- `patch_hf_skip_v2.sh` short-circuits `hf download <repo> --local-dir <dst>` when `dst/config.json` already exists.
- `patch_example_paths.sh` rewrites `/root/<MODEL>` to `/root/models/<MODEL>` and `/root/<DATASET>` to `/root/datasets/<DATASET>` for example scripts (tests already use the prefixed paths).

---

## Update 2026-05-28 — additional r3 image validation (`vime-vllm-r3:test` + PR #48 + PR #55)

Ran another sweep of `tests/test_*.py` on `vime-vllm-r3:test` (vllm nightly w/ PR #39568) on top of PR #48 (3f93600..fd12344) merged on `origin/main` + PR #55 IPC destructure (`_, ipc_args = reduce_tensor(weight)`). Local source patches applied to master worktree:

- `slime/ray/rollout.py:16` — inlined 3 string constants for `GPU_MEMORY_TYPE_*` (drops `from sglang.srt.constants import ...`)
- `slime/backends/megatron_utils/sglang.py` — every sglang import wrapped in `try/except`; `monkey_patch_torch_reductions` falls back to no-op when sglang isn't installed

### Status changes vs the section above

| Test | Old status | New status | Notes |
|---|---|---|---|
| `test_moonlight_16B_A3B_r3.py` | 🔧 needs Bug 3 + Bug 4 patches | ✅ PASS on r3 | Both patches applied (see master patches above). 10-rollout `moonlight-long-r3` long-run also PASS. |
| `test_glm4.7_30B_A3B_pd_mooncake.py` | ⏭️ Mooncake infra blocker | ✅ PASS on r3 | Single-node colocate smoke (`--num-rollout 2`); Mooncake KV store is referenced via sglang disagg backend args but is not strictly required when the test is colocated. Job `raysubmit_SpHjTuBLMiFBybRi` succeeded. |
| `test_qwen3_30B_A3B_r3.py` | ⏭️ blocked on r3 image + FP8 block_n | ❌ image is fine but trips PR #49 R3 routing (Bug 22) | Job FAIL; assertion fires inside `actor.train` on `train_rollout_logprob_abs_diff` / `grad_norm` path. Block reason changed from infra → R3-path-specific. |
| `test_qwen3_4B_ckpt.py` | ❌ `Cannot merge 81 vs 52` | ❌ `Cannot merge 81 vs 79` | Same failure mode; the second number changed (52 → 79) after PR #50, suggesting the underlying shape calculation moved but is still mismatched. Not a PR #18/#40 regression. |
| `test_qwen3_0.6B_parallel_check.py` | ✅ PASS after PR #48 patches | ❌ on r3 image: `AssertionError: grad norm mismatch: 0.502 != 0.422` | **Regression on r3 image when R3 routing replay enabled.** Step 0 completes (lp_diff=0.069, IPC OK) but the cross-config grad-norm consistency assertion at `model.py:795` fails. Worth investigating whether R3 routing replay introduces non-determinism that breaks parallel_check. |

### New PASS entries (not in the table above)

- ✅ `test_qwen3_4B_ppo.py` long-run — 15 rollouts on r3, IPC weight sync stable end-to-end
- ✅ `test_moonlight_16B_A3B.py` long-run — 10 rollouts on r3, MoE IPC stable
- ✅ `test_mimo_7B_mtp_only_grad.py` long-run — 10 rollouts on r3, MTP path stable
- ✅ `test_qwen3_4B_ppo_train_critic_only.py` — 8 GPU on r3 (PR #40 section above had it as ✅ on PR #48 base, confirmed again on r3)

### r3 image systemic incompatibilities (new finding)

The `vime-vllm-r3:test` image is **not a drop-in replacement** for `inferactinc/public:vime-vllm-cu129-latest`. Tests that pass on cu129 fail on r3 because:

1. **`megatron-bridge` package is absent on r3** → any test using `--megatron-to-hf-mode bridge` fails with `ModuleNotFoundError: No module named 'megatron.bridge.models.qwen.qwen_provider'`. Tests affected on r3:
   - `test_qwen2.5_0.5B_debug_rollout_then_train.py`
   - `test_qwen2.5_0.5B_ppo_critic_only_short.py`
   - `test_sglang_config_mixed_offload.py` / `test_sglang_config_mixed_offload_ft.py`
   - `test_qwen2.5_0.5B_sglang_config.py`

2. **sglang is absent on r3** → tests that genuinely require sglang (PD-disagg backends, teacher servers) fail at import / launch:
   - `test_qwen3.6_35B_A3B_pd_mooncake.py` — uses `--sglang-disaggregation-transfer-backend mooncake`; r3 has no sglang to honor it
   - `test_qwen2.5_0.5B_opd_sglang.py` — launches an sglang teacher server; exits with code 1

PR #55's IPC 15-tuple format mandates the r3 image (cu129 vllm 0.21.0 cannot unpack the new format — `func, args = handle` raises `ValueError: too many values to unpack (expected 2)`). So PR #55 + r3 is a hard pairing. Options:

- (a) CI matrix runs both cu129 (for sglang/bridge paths) **and** r3 (for the PR #48/#55 IPC paths) — accept that no single image covers everything.
- (b) Add `sglang` + `megatron-bridge` to the r3 image so it becomes a superset of cu129.

### Open follow-up beyond PR #48

- Bug 16: `test_qwen3_4B_ckpt.py` Megatron checkpoint merge — orthogonal to IPC, separate PR.
- Bug 22: PR #49 R3 routing-replay on GSPO+MoE — triggers AssertionError on `test_qwen3_30B_A3B_r3.py` and `test_qwen3_0.6B_parallel_check.py`. Needs PR #49 owner.
- Bug 28: `megatron-bridge` missing on r3 — packaging concern, not slime.


tests/unit/backends/vllm_utils/test_arguments.py
```diff
@@ -145,7 +145,7 @@ def _ns(**overrides):
         vllm_data_parallel_size=1,
         vllm_pipeline_parallel_size=1,
         rollout_num_gpus_per_engine=4,
-        router_ip=None,
+        vllm_router_ip=None,
```
CalvinXKY, line 148: If we want to move to `--vllm-router-*`, could we change `arguments.py` + all call sites in the same PR so docs/tests/runtime stay consistent?

slime/rollout/on_policy_distillation.py
```diff
@@ -6,50 +56,77 @@
...
+        "max_tokens": 1,
+        "temperature": 0,
+        "prompt_logprobs": 1,
+        "logprobs": 0,
+        "skip_special_tokens": False,
     }
 
     if sample.multimodal_inputs and sample.multimodal_inputs.get("images"):
         image_data = sample.multimodal_inputs["images"]
+        # vLLM accepts multimodal inputs via the chat completions endpoint
```
aoshen02: problem

slime/rollout/on_policy_distillation.py
```diff
@@ -6,50 +56,77 @@
...
     This function:
-    1. Extracts teacher log-probs from the reward response (which contains sglang's logprob output)
-    2. Trims them to match the response length
-    3. Stores them in sample.teacher_log_probs for OPD KL penalty computation
-    4. Returns scalar rewards (0.0 for pure distillation) compatible with GRPO/PPO
-
-    Note: The reward_func calls the teacher server which returns token-level log-probs.
-    For pure on-policy distillation without task rewards, we return 0.0 for each sample.
-    The actual learning signal comes from the OPD KL penalty applied in compute_advantages_and_returns.
+    1. Extracts teacher log-probs from the teacher's `/v1/completions`
```
aoshen02: 怎么还在v1

tests/unit/backends/vllm_utils/test_arguments.py
aoshen02: I think in this case then we will need to merge these two into one pr, which is also ok after I review it.

PR #42, opened 2026-05-27
Rename slime references to vime

aoshen02:
## Summary
- Rename the core Python package from `slime` to `vime` and plugin package from `slime_plugins` to `vime_plugins`.
- Update docs, scripts, CI config, examples, tests, and patch paths from slime naming to vime naming.
- Rename `docker/npu_patch/slime.patch` to `docker/npu_patch/vime.patch`.
- Preserve upstream `slime` attribution for inherited training-stack behavior, historical Z.ai/slime context, upstream docs, and acknowledgements.
- Fix mechanically rewritten `THUDM/vime` links to point at `vllm-project/vime`, while restoring upstream slime links where they refer to slime history or upstream PRs.

## Validation
- `git diff --check`
- `python -m compileall -q train.py train_async.py vime vime_plugins tests`
- `python - <<'PY' ... import vime; import vime_plugins ... PY`
- `PYTHONPATH=. pytest -q tests/unit tests/plugin_contracts` currently fails in this environment due missing/partial external deps (`megatron`, vLLM/transformers import), plus an existing plugin contract marker mismatch that is already present on the base branch.


CalvinXKY:
The renaming operation should be done after the smoke test is set up.

gcanlin (approved):
LGTM. I have run the UTs and they can pass. And I'm running Qwen3-4B train with TP4 + TP4. But it hasn't run completely.

Image

gcanlin:
image

PR #47, opened 2026-05-27
fix(vllm): unify weight-transfer HTTP handling

aoshen02:
Follow-up to #22 and rebased over #49 / current `main`. PR #49 merged the R3 routing-replay work and partially centralized the vLLM weight-transfer timeout path, but current `main` still had two issues in the vLLM control plane:

- Response parsing failures were converted to `{"ok": True, "raw": ...}` in several endpoints, which can hide bad/plain-text responses as success.
- `init_weight_transfer_engine` / `init_weights_update_group` still used a separate timeout path from the rest of the weight-transfer calls.

## Why this is not duplicating #49

#49 was a routing replay feature PR. It did touch `vllm_engine.py`, and it left `start_weight_update` / `finish_weight_update` using a timeout helper, but it did not make response parsing fail loud and did not move all weight-transfer control-plane calls to a config-backed timeout. This PR is now scoped to those remaining control-plane fixes.

## Changes

- Replace the old fallback parser with a slime-style `_response_json(response)` helper:
  - `raise_for_status()` first.
  - Attach `response.text` to HTTP errors with `Exception.add_note`, matching the SGLang engine pattern.
  - Parse JSON directly and let invalid JSON fail instead of returning `ok: True`.
- Add `--vllm-weight-transfer-timeout-sec` (default `900.0`) as the vime config knob for vLLM weight-transfer HTTP control-plane calls.
- Exclude that vime-only config knob from forwarding to `vllm serve`.
- Route `update_weights`, `start_weight_update`, `finish_weight_update`, `init_weight_transfer_engine`, and `init_weights_update_group` through the same timeout helper.
- Remove the old `SLIME_VLLM_WEIGHT_TRANSFER_*_TIMEOUT_SEC` env-var path from this code.

## Tests

```bash
uv run --active python -m pytest tests/unit/backends/vllm_utils/test_vllm_engine.py tests/unit/backends/vllm_utils/test_arguments.py -q
# 66 passed, 2 warnings in 85.43s

uv run --active pre-commit run --files slime/backends/vllm_utils/arguments.py slime/backends/vllm_utils/vllm_engine.py tests/unit/backends/vllm_utils/test_arguments.py tests/unit/backends/vllm_utils/test_vllm_engine.py
# all selected hooks passed
```

## AI assistance

AI assistance was used to inspect related PRs/reference implementations and update this PR. The human submitter should review every changed line and validate the behavior before merge.

### DIFF
--- a/slime/backends/vllm_utils/vllm_engine.py
+++ b/slime/backends/vllm_utils/vllm_engine.py
@@ -32,6 +32,15 @@ def _normalize_vllm_wake_tags(tags: list[str] | None) -> list[str] | None:
     return normalized or None
 
 
+def _response_json(response: requests.Response) -> dict:
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError as e:
+        e.add_note(f"{response.text=}")
+        raise
+    return response.json()
+
+
 def get_base_gpu_id(args, rank):
     """First local GPU index on this node for rollout engine *rank* (colocate vs actor[/critic]-offset layout)."""
     num_gpus = min(args.num_gpus_per_node, args.rollout_num_gpus_per_engine)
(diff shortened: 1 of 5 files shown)


CalvinXKY:
LGTM. Rebase to solve the conflicts

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -32,6 +32,16 @@ def _normalize_vllm_wake_tags(tags: list[str] | None) -> list[str] | None:
     return normalized or None
 
 
+def _response_json(response: requests.Response) -> dict:
+    """Return a JSON response using the same fail-loud behavior as SGLangEngine."""
```
CalvinXKY: remove “as SGLangEngine.”

CalvinXKY:
LGTM

PR #48, opened 2026-05-27
refactor(weight-sync): align IPC RPC contract with slime (single-RPC version-with-data)

aoshen02:
## Summary

Follow-up to #45. Centralises ``_weight_version`` bookkeeping inside the data-carrying RPC so the IPC and distributed paths share the same contract slime uses, and removes two dead methods that were left behind by the PR #18 → PR #22 transition.

## Why

After #22 introduced the colocated CUDA IPC path, vime had three ``update_weights*`` entry points whose version bookkeeping was inconsistent:

| Entry | Path | Records `_weight_version`? |
|------|------|---|
| `update_weights_from_distributed` | NCCL | ✅ in-RPC |
| `update_weights` (called from `IPCWeightTransferEngine.trainer_send_weights`) | IPC | ❌ — vLLM's `IPCWeightTransferUpdateInfo` schema has no `weight_version` field |
| `update_weights_from_tensor` (PR #18 legacy) | — | ✅ in-RPC, but no callers |

The IPC gap caused the `Weight version mismatch! Engine: /root/models/<...>, Updater: N` failure under `--ci-test`. #45 patched it by piggybacking `weight_version` onto `finish_weight_update`, which fixed the bug but coupled version bookkeeping to a lifecycle hook instead of the data RPC. Any future IPC-style entry would face the same "remember to also write `_weight_version`" trap that PR #22 originally fell into.

slime avoids this trap by design: both IPC and distributed paths call `engine.update_weights_from_tensor.remote(..., weight_version=N)` (see ``slime/backends/sglang_utils/sglang_engine.py:update_weights_from_tensor``) — same RPC name, version travels with the data.

## Changes

1. **`vllm_engine.update_weights_from_tensor`** rewritten as the IPC entry point. Signature ``(update_info: dict, weight_version: str | None, flush_cache: bool)``; payload carries vLLM's ``IPCWeightTransferUpdateInfo`` (names / dtype_names / shapes / ipc_handles). Records ``_weight_version`` only **after** the POST succeeds — same post-POST ordering as ``update_weights_from_distributed`` and the review note on #45.
2. **Delete `vllm_engine.update_weights`** — was the entry for vLLM upstream's ``IPCWeightTransferEngine.trainer_send_weights``; no longer used after step 5.
3. **Delete `vllm_engine._run_vllm_weight_update`** — dead helper; only the old `update_weights_from_tensor` body called it.
4. **Revert `finish_weight_update` to a stateless POST** — undoes the `weight_version` kwarg added in #45 since the version now lives in the data RPC.
5. **`update_weight_from_tensor._send_hf_chunk_via_ipc`** stops using `IPCWeightTransferEngine.trainer_send_weights`. Both `slot_size == 1` and `slot_size > 1` branches now call `engine.update_weights_from_tensor.remote(update_info=..., weight_version=str(self.weight_version))` directly. vime keeps using `reduce_tensor` from `torch.multiprocessing.reductions` for IPC handle creation (via `_build_ipc_update_info_from_named_tensors`); we only own the dispatch.

## Why not just keep #45

`#45` worked but the wire shape did not match slime, and the bookkeeping write lived on a lifecycle hook (`finish_weight_update`) rather than the data RPC. Anyone adding a future IPC-flavoured entry point would face the same trap that PR #22 fell into. Centralising the write inside the data RPC removes the trap.

## Tests

- `RecordingVLLMEngine` learns `update_weights_from_tensor` so engine RPC call recording stays complete.
- Renamed `test_trainer_send_weights_uses_single_llm_handle_per_rank` → `test_send_via_ipc_dispatches_update_weights_from_tensor_with_version`. Asserts the new RPC name + kwargs and that `finish_weight_update` is stateless (no kwargs).
- Added `test_update_weights_from_tensor_posts_ipc_update_info_and_records_version`: asserts `ipc_handles` get cloudpickle'd into `ipc_handles_pickled`, metadata fields pass through, `_weight_version` advances on POST success.
- Added `test_update_weights_from_tensor_does_not_advance_version_on_failure`: POST failure must not advance `_weight_version` (otherwise a retry would skip the resync).

Pre-existing test failures in `tests/unit/backends/vllm_utils/test_vllm_engine.py` (`_weight_transfer_http_timeout`, `_response_json_or_fallback`, `server_host`) are unchanged from main: main HEAD shows **8 failed / 24 passed**, this PR shows **8 failed / 26 passed** (the two new tests). Those are scope for a separate cleanup.

## Test plan

- [x] Unit: `tests/unit/backends/vllm_utils/test_vllm_engine.py` — 26 passed (was 24 on main), 8 pre-existing failures unchanged
- [x] Unit: `tests/unit/backends/megatron_utils/update_weight/test_update_weight_from_tensor.py` — all passing
- [ ] e2e: rerun the colocate+IPC suite once CI picks it up (`tests/test_qwen2.5_0.5B_short.py`, `tests/test_qwen2.5_0.5B_ppo_critic_only_short.py`, `tests/test_qwen3_4B_ppo.py`)

### DIFF
--- a/slime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
+++ b/slime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
@@ -11,11 +11,23 @@ https://docs.vllm.ai/en/stable/examples/rl/rlhf_ipc/
 
 The flow for colocated engines:
 1. Megatron params → HF conversion (via HfWeightIteratorBase)
-2. All trainer ranks call ``IPCWeightTransferEngine.trainer_send_weights()``
-   with ``send_mode="ray"`` pointing at the colocated vLLM engine actor on the
-   same GPU slot.  Each rank creates a CUDA IPC handle for its GPU; the engine
-   collects all handles via ``_all_gather_and_merge_handles`` so every vLLM
-   worker can pick the handle belonging to its physical GPU UUID.
+2. Each trainer rank builds vLLM-protocol IPC handles for its GPU
+   (``reduce_tensor`` from ``torch.multiprocessing.reductions``) and dispatches
+   them via ``engine.update_weights_from_tensor.remote(update_info=..., weight_version=...)``.
+   When ``rollout_num_gpus_per_engine > 1`` (vLLM TP within the engine), the
+   coordinator rank gathers and merges per-rank handles into one RPC so every
+   vLLM worker can pick the handle for its physical GPU UUID. The RPC name
+   ``update_weights_from_tensor`` and the ``weight_version`` kwarg match slime's
+   IPC entry point (``sglang_engine.update_weights_from_tensor``); only the
+   on-the-wire payload differs (vLLM IPCWeightTransferUpdateInfo dict vs
+   SGLang serialized_named_tensors).
+
+We deliberately do not use ``IPCWeightTransferEngine.trainer_send_weights``:
+its hard-coded ``llm_handle.update_weights.remote(...)`` call would force the
+RPC contract to be the vLLM one and would drop ``weight_version`` on the floor
+(ci_test's engine-vs-updater check requires the engine to track the version).
+Reusing only ``reduce_tensor`` for handle creation lets vime keep the RPC
+contract aligned with slime.
 
 For non-colocated overflow engines the existing NCCL distributed broadcast
 (``update_weights_from_distributed``) is used unchanged.
@@ -122,18 +134,10 @@ def _merge_ipc_update_infos(infos: Sequence[dict[str, list]]) -> dict[str, list]
 
 
 class UpdateWeightFromTensor:
-    """
-    Update colocated vLLM engines from tensors via CUDA IPC (Ray send mode).
-
-    Colocated path:
-        Megatron weights → HF conversion → CUDA IPC to vLLM engine actors via
-        ``IPCWeightTransferEngine.trainer_send_weights(send_mode="ray")``.
-        Each trainer rank sends to the colocated engine on its GPU slot.
-
-    Distributed overflow path (optional):
-        Falls back to NCCL distributed broadcast via
-        ``update_weights_from_distributed`` for engines whose GPUs lie outside
-        the actor GPU range.
+    """Update colocated vLLM engines via CUDA IPC, with NCCL fallback for
+    non-colocated overflow engines. See the module docstring for the
+    high-level design (why we dispatch via ``update_weights_from_tensor``
+    directly instead of vLLM's ``trainer_send_weights``).
 
     Engine lifecycle per ``update_weights`` call::
 
@@ -142,7 +146,7 @@ class UpdateWeightFromTensor:
         init_weight_transfer_engine                      (rank 0, colocated, first call only)
         start_weight_update                              (coordinator rank per engine only)
         [for each HF chunk]
-          trainer_send_weights                           (each rank mapped to _ipc_engine)
+          update_weights_from_tensor                     (per-rank or coordinator-merged)
           update_weights_from_distributed                (src rank, distributed)
           barrier                                        (all ranks)
         finish_weight_update                             (coordinator rank per engine only)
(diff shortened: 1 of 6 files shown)


aoshen02:
CI baseline check: confirmed the 5 `e2e-test-plugin-contracts` failures (`test_megatron_argument_validation`, `test_plugin_path_loading_contracts`, `test_plugin_runtime_hook_contracts`, `test_plugin_rollout_contracts`, `test_plugin_generate_contracts`) and the `Run pre-commit` failure are **pre-existing on main** — same failures reproduce locally on `origin/main` (HEAD `68ca256`) without any of this PR's changes applied. Not introduced by this PR.

Concretely for `test_plugin_runtime_hook_contracts.py::test_runtime_hook_callsite_is_stable[rollout_data_postprocess]` — the contract expects `self.rollout_data_postprocess(self.args)` in `slime/backends/megatron_utils/actor.py`, but main's actor.py doesn't contain that string (callsite drift). This PR doesn't touch `actor.py` at all.

Local unit-test result: 29 passed (24 → 29 = +5 new tests for this PR), 8 pre-existing failed (same 8 failing on `origin/main`).

slime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -237,8 +243,23 @@ def connect_rollout_engines(
...
+        self._ipc_slot_group = None
+        rank_for_slot = dist.get_rank()
         colocate_gpu_offsets = engine_gpu_offsets[:colocate_engine_nums]
         colocate_gpu_counts = engine_gpu_counts[:colocate_engine_nums]
+        # First pass: create per-slot process groups collectively (every rank must call new_group).
+        for i in range(colocate_engine_nums):
+            slot_start = colocate_gpu_offsets[i]
+            slot_end = slot_start + colocate_gpu_counts[i]
+            slot_ranks = list(range(slot_start, slot_end))
+            grp = dist.new_group(ranks=slot_ranks)
```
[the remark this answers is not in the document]
aoshen02: not a problem

knlnguyen1802: The `_ipc_slot_group` is created here without destroy anywhere. There is potential of memory  leak if  `connect_rollout_engines` is called many times even though it's a rare case. Consider fix or add a note on this

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -624,37 +594,39 @@ def health_generate(self, timeout: float = 5.0) -> bool:
...
-            self._weight_version = str(weight_version)
+        payload: dict = {"names": names, "dtype_names": dtype_names, "shapes": shapes}
+        if ipc_handles is not None:
+            import base64
+
+            import cloudpickle
+
+            payload["ipc_handles_pickled"] = base64.b64encode(cloudpickle.dumps(ipc_handles)).decode("utf-8")
         if flush_cache:
             self.flush_cache()
```
knlnguyen1802, line 623: It looks like this is never call under current code, same issue with `update_weight_from_distributed` can fix it in other PR

SamitHuang:
I've addressed the review feedback:
- Explicitly specified the `gloo` backend for the process group creation to support `all_gather_object` calls.
- Added a unit test (`test_send_via_ipc_dispatches_update_weights_from_tensor_coordinator_multi_gpu`) to cover the coordinator path when `slot_size > 1`.

knlnguyen1802:
> I've addressed the review feedback:
> 
> * Explicitly specified the `gloo` backend for the process group creation to support `all_gather_object` calls.
> * Added a unit test (`test_send_via_ipc_dispatches_update_weights_from_tensor_coordinator_multi_gpu`) to cover the coordinator path when `slot_size > 1`.

@SamitHuang  Please fix DCO

SamitHuang (approved):
lgtm overall

slime/backends/vllm_utils/vllm_engine.py
aoshen02: We might need to refactor this, not sure is it OK to keep self.node_rank = 0

CalvinXKY:
LGTM

PR #34, opened 2026-05-26
[Feature]Routing replay (R3) for vLLM rollout  (vLLM v0.21.0)

CalvinXKY:
## Purpose

Implements **Phase 1** of [RFC #32](https://github.com/vllm-project/vime/issues/32): MoE **routing replay** (`--use-rollout-routing-replay`) on the **vLLM rollout** path, pinned to **vLLM 0.21.0**.

SGLang already fills `sample.rollout_routed_experts` with shape `(len(tokens) - 1, num_layers, moe_router_topk)`. vLLM 0.21.0 records routing internally but `/inference/v1/generate` did not expose it, and `vllm_rollout` only handled base64 npy on `choices[].routed_experts`.

**Follow-up (not this PR):** colocate E2E aligned with `tests/test_qwen3_30B_A3B_r3.py`; upgrade to latest vLLM (RFC Phase 2).

## What's included

- **`docker/patch/latest/vllm.patch`** — add `prompt_routed_experts` + `choices[].routed_experts` on `/inference/v1/generate`; bake in `docker/Dockerfile`; docs in `docker/README.md`.
- **`slime/rollout/vllm_rollout.py`** — merge prompt + gen routing (nested list or base64 npy); assign Megatron-shaped `rollout_routed_experts`.
- **`slime/backends/vllm_utils/vllm_engine.py`** — when R3 enabled: `--enable-return-routed-experts`, `--no-async-scheduling` (unless overridden); `--enable-expert-parallel` + `linear` placement only if `--vllm-enable-expert-parallel`.
- **Tests** — `tests/unit/rollout/test_vllm_rollout.py`; shape checks in `tests/test_vllm_generate_endpoint.py` (R3 case).

## Test plan

**Model:** Qwen3-30B-A3B (`num_layers=48`, `moe_router_topk=8`)

### Unit

```bash
pytest tests/unit/rollout/test_vllm_rollout.py -k routed_experts
```

### Integration (GPU + model)

```bash
pytest tests/test_vllm_generate_endpoint.py -k r3
```

### Manual — decoupled (Gate B, per RFC #32)

1. Image with patch applied (`docker build -f docker/Dockerfile ...` or runtime patch per `docker/README.md`).
2. Rollout flags: `--use-rollout-routing-replay --vllm-enable-expert-parallel` (MoE multi-GPU), no `--colocate`.
3. Confirm `/inference/v1/generate` returns `prompt_routed_experts` and `choices[].routed_experts`.
[...]

Issue #32, opened 2026-05-26
[RFC] Routing Replay (R3) for vLLM Rollout Engine


CalvinXKY:
**Status:** In progress  
**Target model:** Qwen3-30B-A3B (48 MoE layers, `moe_router_topk=8`)  
**Rollout backend:** vLLM (via slime `vllm_rollout`)

---

## Summary

**Routing Replay (R3)** records MoE expert routing decisions during rollout and replays them during Megatron training so the training forward pass uses the same expert assignments as inference.

SGLang already supports R3 end-to-end (`--enable-return-routed-experts` + `meta_info.routed_experts`). vLLM exposes routing internally when launched with `--enable-return-routed-experts`, but the token-only `/inference/v1/generate` path used by slime’s `vllm_rollout` did not surface those fields on the HTTP response, and the client did not merge prompt vs. generation segments into the tensor shape Megatron expects.

This RFC defines a **phased plan** to bring vLLM rollout to R3 parity with SGLang, starting with the **currently deployed vLLM 0.21.0** stack, then upgrading to newer vLLM releases.

---

## Phased Roadmap

| Phase | Scope | Goal |
|-------|--------|------|
| **Phase 1 (current)** | **vLLM 0.21.0** | Patch HTTP, align Megatron tensor contract, update vime client/engine launcher, validate decoupled then colocate |
| **Phase 2 (follow-up)** | **Latest vLLM** | Rebase or drop patch, re-validate API payload & launch flags, extend CI matrix |

Phase 1 is a **version-pinned adapter**. Phase 2 is a separate upgrade track once Phase 1 is stable in training.

---

## Background

When `--use-rollout-routing-replay` is enabled:

1. The rollout engine returns per-token expert indices for every MoE layer.
2. `vllm_rollout` / `sglang_rollout` populate `sample.rollout_routed_experts`.
3. Megatron actor consumes `rollout_routed_experts` during the training step.

Megatron requires:

```text
rollout_routed_experts.shape == (len(sample.tokens) - 1, num_layers, moe_router_topk)
dtype: int32
```

For Qwen3-30B-A3B: `(seq_len - 1, 48, 8)`.

**SGLang (reference):** single base64 int32 buffer in `meta_info.routed_experts`, reshaped directly.

**vLLM 0.21.0 (engine internal):** routing split into:

| Segment | HTTP field | Role |
|---------|------------|------|
| Prompt | `prompt_routed_experts` (response top-level) | Prompt-phase routing |
[...]

PR #49, opened 2026-05-27
[Feature] Routing replay (R3) for vLLM rollout (vLLM 0.22+, /inference/v1/generate)

CalvinXKY:
## Purpose

Implements **RFC #32 Phase 2** / follow-up to [#34](https://github.com/vllm-project/vime/pull/34): MoE **routing replay** (`--use-rollout-routing-replay`) on the **vLLM rollout** path using **vLLM ≥ 0.22** native support ([vllm#39568](https://github.com/vllm-project/vllm/pull/39568)).

SGLang already fills `sample.rollout_routed_experts` with shape `(len(tokens) - 1, num_layers, moe_router_topk)`. On vLLM 0.21.x, routing required a **site-packages patch** and/or **`/v1/completions`** workarounds. From vLLM 0.22+, **`/inference/v1/generate`** returns routed experts (often as a single base64 npy buffer on `choices[].routed_experts`).

This PR **drops the disagg patch path** and standardizes rollout + engine smoke checks on **`/inference/v1/generate`**.

**Supersedes the approach in #34** (v0.21.0 + `docker/patch/latest/vllm.patch` + completions fallback).

### Rollout (`slime/rollout/vllm_rollout.py`)

- R3 text generation uses **`POST /inference/v1/generate`** only (removed `/v1/completions` branch).
- **`_merge_generate_routed_experts`**: merge split or **single-buffer** routing; align to `len(tokens) - 1`.

### Engine (`slime/backends/vllm_utils/vllm_engine.py`)

- When `use_rollout_routing_replay`: inject **`--enable-return-routed-experts`** only (no 0.21.x `--no-async-scheduling` / `--no-enable-prefix-caching` guards).
- **`_verify_generate_routed_experts`**: startup smoke on `/inference/v1/generate`.

### Tests

- `tests/unit/rollout/test_vllm_rollout.py` — merge/apply routing, **single-buffer** case.
- `tests/unit/backends/vllm_utils/test_vllm_engine.py` — `_verify_generate_routed_experts`.
- `tests/test_vllm_generate_endpoint.py` — `qwen3-30b-a3b-r3` integration.

## Requirements

| Item | Notes |
|------|--------|
| vLLM | **≥ 0.22.0** (or dev build with #39568) |
| MoE multi-GPU | `--vllm-enable-expert-parallel` |
| R3 | `--use-rollout-routing-replay` |

## Verification logs

Environment: **8×A100**, container `vime_v22`, Qwen3-30B-A3B, train/rollout **4+4**, vLLM `0.21.1rc1.dev38` (includes #39568 generate routing). No `Split sizes` / missing-routing errors observed.

### 1) Local unit tests

```text
pytest tests/unit/rollout/test_vllm_rollout.py tests/unit/backends/vllm_utils/test_vllm_engine.py -q
# 87 passed in 3.08s

pre-commit run --all-files
# all hooks Passed (ruff, black, isort, …)
```

### 2) Full verify — HTTP + rollout E2E

**Phase A — standalone vLLM HTTP smoke (`:8000`)**

```text
[routing-replay-full] Starting standalone vLLM on http://127.0.0.1:8000 (GPUs=0,1,2,3, tp=4) ...
[routing-replay-full] HTTP model id: /data/nfs_87/model/Qwen3-30B-A3B
[vllm-generate-routing] POST http://127.0.0.1:8000/inference/v1/generate
  routing rows=12 layers=48 top_k=8
OK: /inference/v1/generate routing replay fields present.
[routing-replay-full] HTTP phase PASS
```

**Phase B — `debug-rollout-only` + `.pt` shape check**

```text
(VLLMEngine) non-default args: {..., 'enable_return_routed_experts': True, 'enable_expert_parallel': True, ...}
(EngineCore) RoutedExpertsManager CPU buffer: 0.63 GB (slots=1642320, layers=48, top_k=8, dtype=uint8)
(APIServer) Route: /inference/v1/generate, Methods: POST
(APIServer) "POST /inference/v1/generate HTTP/1.1" 200 OK

  sample[0]: OK shape=(322, 48, 8) tokens=323 response_length=128
  sample[1]: OK shape=(265, 48, 8) tokens=266 response_length=128
PASS: 2 sample(s) satisfy routing replay shape contract.
[routing-replay-verify] Rollout path PASS
```

### 3) Production-shaped training — 4+4 R3 smoke

```text
(VLLMEngine) non-default args: {..., 'enable_return_routed_experts': True,
  'tensor_parallel_size': 4, 'enable_expert_parallel': True, ...}
(Worker) Initializing routed experts capturer, enable_return_routed_experts: True
(APIServer) Route: /inference/v1/generate, Methods: POST
(APIServer) "POST /inference/v1/generate HTTP/1.1" 200 OK

(MegatronTrainRayActor) torch.from_numpy(r) for r in rollout_data["rollout_routed_experts"]
(RolloutManager) perf 0: {..., 'perf/tokens_per_gpu_per_sec': 305.09, ...}
(MegatronTrainRayActor) perf 0: {..., 'perf/actor_train_tok_per_s': 687.56, 'perf/step_time': 223.28, ...}
```

Training completed multiple rollout→train steps without routing-shape failures.

### Unit

```bash
pytest tests/unit/rollout/test_vllm_rollout.py -k "routed or merge_generate"
pytest tests/unit/backends/vllm_utils/test_vllm_engine.py -k verify_generate
```

### Integration (GPU)

```bash
pytest tests/test_vllm_generate_endpoint.py -k r3
```

### Manual

1. vLLM ≥ 0.22 + `--enable-return-routed-experts` (no patch).
2. `--use-rollout-routing-replay --vllm-enable-expert-parallel`, 4+4 layout.
3. Confirm `rollout_routed_experts` shape `(len(tokens)-1, num_layers, moe_router_topk)`.

## Out of scope

- Docker base bump to `v0.22.0` image tag.
- Colocate CI parity with `tests/test_qwen3_30B_A3B_r3.py`.
- Removing legacy `docker/patch` tree.

## Related

- #32 (RFC)
- #34 (v0.21 + patch — superseded)
- [vllm#39568](https://github.com/vllm-project/vllm/pull/39568)

Related to #32 (Phase 2).

### DIFF
--- a/slime/backends/vllm_utils/vllm_engine.py
+++ b/slime/backends/vllm_utils/vllm_engine.py
@@ -5,6 +5,10 @@ import logging
 import multiprocessing
 import os
 import time
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import numpy as np
 from urllib.parse import quote
 
 import requests
@@ -308,9 +312,6 @@ def launch_server_process(
     # unless the user already passed --vllm-max-model-len explicitly.
     if args.rollout_max_context_len is not None and getattr(args, "vllm_max_model_len", None) is None:
         cmd += ["--max-model-len", str(args.rollout_max_context_len)]
-    if getattr(args, "use_rollout_routing_replay", False):
-        cmd += ["--enable-return-routed-experts"]
-
     # vime-preferred defaults — must be explicitly forwarded because the vllm-side
     # default would otherwise apply (the generic forwarder skips values that equal
     # action.default).
@@ -335,6 +336,22 @@ def launch_server_process(
             return False
         _, action = entry
         return getattr(args, dest, action.default) != action.default
+
+    # MoE routing replay: routed-experts return + sync scheduling (incompatible with
+    # async scheduling on vLLM 0.21.x). Expert parallel is opt-in via
+    # ``--vllm-enable-expert-parallel``.
+    if getattr(args, "use_rollout_routing_replay", False):
+        cmd += ["--enable-return-routed-experts"]
+        if not _user_overrode("vllm_async_scheduling"):
+            cmd += ["--no-async-scheduling"]
+        # Prefix cache hits skip prefill MoE forwards; routed-experts capture then
+        # only covers decode (~gen_len-1 rows) and prompt_routed_experts is missing.
+        if not _user_overrode("vllm_enable_prefix_caching"):
+            cmd += ["--no-enable-prefix-caching"]
+        if getattr(args, "vllm_enable_expert_parallel", False):
+            cmd += ["--enable-expert-parallel"]
+            if not _user_overrode("vllm_expert_placement_strategy"):
+                cmd += ["--expert-placement-strategy", "linear"]
 
     # 1) gpu_memory_utilization: vllm default 0.92 OOMs in colocate training; vime ships 0.55.
     if _user_overrode("vllm_gpu_memory_utilization"):
@@ -403,6 +420,76 @@ def _redact_cmd_for_log(cmd: list[str]) -> str:
         if isinstance(token, str) and token in _REDACTED_FLAGS:
             redact_next = True
     return " ".join(parts)
+
+
+def _routing_rows_from_http_payload(value: Any) -> np.ndarray | None:
+    """Decode vLLM routed-experts HTTP field (base64 npy or nested list)."""
+    import base64
+    import io
+
+    import numpy as np
+
+    if value is None:
+        return None
+    if isinstance(value, str):
+        return np.load(io.BytesIO(base64.b64decode(value)), allow_pickle=False)
+    if isinstance(value, list):
+        return np.asarray(value, dtype=np.int32)
+    return None
+
+
+def _verify_generate_routed_experts(base_url: str, model: str, timeout_s: float = 120.0) -> None:
+    """Smoke-check MoE routing via ``/v1/completions`` (matches R3 rollout on vLLM 0.21.x)."""
+    import numpy as np
+
+    base = base_url.rstrip("/")
+    payload = {
+        "model": model,
+        "prompt": "Routing replay smoke test.",
+        "max_tokens": 8,
+        "temperature": 0.0,
+        "logprobs": 1,
+        "return_token_ids": True,
+        "stream": False,
+    }
+    response = requests.post(f"{base}/v1/completions", json=payload, timeout=timeout_s)
+    response.raise_for_status()
+    body = response.json()
+    choice = (body.get("choices") or [{}])[0]
+    pre = _routing_rows_from_http_payload(body.get("prompt_routed_experts"))
+    gen = _routing_rows_from_http_payload(choice.get("routed_experts"))
+    if pre is None and gen is None:
+        raise RuntimeError(
+            "vLLM /v1/completions returned no routed-experts fields. "
+            "Ensure the server was started with --enable-return-routed-experts "
+            "(use_rollout_routing_replay)."
+        )
+    if pre is None or gen is None:
+        raise RuntimeError(
+            "vLLM /v1/completions must return both prompt_routed_experts and "
+            "choices[].routed_experts for routing replay. "
+            "Ensure --enable-return-routed-experts and --no-async-scheduling on the vLLM cmdline."
+        )
+
+    out_ids = choice.get("token_ids") or []
+    usage = body.get("usage") or {}
+    num_prompt = int(usage.get("prompt_tokens") or 0)
+    num_gen = int(usage.get("completion_tokens") or len(out_ids))
+    expected_rows = num_prompt + num_gen - 1 if num_prompt > 0 and num_gen > 0 else 0
+
+    merged = np.concatenate([pre, gen], axis=0)
+    n_rows = int(merged.shape[0])
+    if expected_rows > 0 and n_rows not in (expected_rows, expected_rows + 1):
+        raise RuntimeError(
+            f"vLLM routing replay smoke check: merged routing rows {n_rows} != "
+            f"expected {expected_rows} (prompt+gen len(tokens)-1). "
+            "Ensure --enable-return-routed-experts and --no-async-scheduling."
+        )
+    logger.info(
+        "vLLM routing replay smoke check OK (/v1/completions): prompt+gen routing rows=%s " "(expected %s)",
+        n_rows,
+        expected_rows,
+    )
 
 
 def _wait_server_healthy(base_url: str, process: multiprocessing.Process | None, timeout_s: float = 300.0) -> None:
@@ -559,7 +646,10 @@ class VLLMEngine(RayActor):
             visible_devices=visible_devices,
             model_path=self.model_path,
         )
-        _wait_server_healthy(self._http_base(), process=self.process)
+        base = self._http_base()
+        _wait_server_healthy(base, process=self.process)
+        if getattr(self.args, "use_rollout_routing_replay", False):
+            _verify_generate_routed_experts(base, self.model_path)
 
     def _post_json(self, endpoint: str, payload: dict, timeout: float) -> requests.Response:
         url = f"{self._http_base()}/{endpoint.lstrip('/')}"
--- a/slime/backends/vllm_utils/vllm_engine.py
+++ b/slime/backends/vllm_utils/vllm_engine.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import ipaddress
+import json
 import logging
 import multiprocessing
 import os
@@ -337,21 +338,9 @@ def launch_server_process(
         _, action = entry
         return getattr(args, dest, action.default) != action.default
 
-    # MoE routing replay: routed-experts return + sync scheduling (incompatible with
-    # async scheduling on vLLM 0.21.x). Expert parallel is opt-in via
-    # ``--vllm-enable-expert-parallel``.
+    # MoE routing replay (vLLM 0.22+, PR #39568): routed experts on ``/inference/v1/generate``.
     if getattr(args, "use_rollout_routing_replay", False):
         cmd += ["--enable-return-routed-experts"]
-        if not _user_overrode("vllm_async_scheduling"):
-            cmd += ["--no-async-scheduling"]
-        # Prefix cache hits skip prefill MoE forwards; routed-experts capture then
-        # only covers decode (~gen_len-1 rows) and prompt_routed_experts is missing.
-        if not _user_overrode("vllm_enable_prefix_caching"):
-            cmd += ["--no-enable-prefix-caching"]
-        if getattr(args, "vllm_enable_expert_parallel", False):
-            cmd += ["--enable-expert-parallel"]
-            if not _user_overrode("vllm_expert_placement_strategy"):
-                cmd += ["--expert-placement-strategy", "linear"]
 
     # 1) gpu_memory_utilization: vllm default 0.92 OOMs in colocate training; vime ships 0.55.
     if _user_overrode("vllm_gpu_memory_utilization"):
@@ -422,6 +411,17 @@ def _redact_cmd_for_log(cmd: list[str]) -> str:
     return " ".join(parts)
 
 
+def _response_json_or_fallback(response) -> dict:
+    """Parse JSON from an HTTP response, returning a structured error dict on failure."""
+    try:
+        data = response.json()
+    except (ValueError, json.JSONDecodeError):
+        return {"ok": False, "error": "Invalid JSON response", "raw": getattr(response, "text", "")}
+    if not isinstance(data, dict):
+        return {"ok": False, "error": "Response is not a dictionary", "data": data}
+    return data
+
+
 def _routing_rows_from_http_payload(value: Any) -> np.ndarray | None:
     """Decode vLLM routed-experts HTTP field (base64 npy or nested list)."""
     import base64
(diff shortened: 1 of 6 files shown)


CalvinXKY:
# E2E test：

qwen3-30B-A3B-4t4i-r3-300step.sh

image

slime/ray/rollout.py
```diff
@@ -774,7 +774,17 @@ def _convert_samples_to_train_data(self, samples: list[Sample] | list[list[Sampl
...
+            routed = [sample.rollout_routed_experts for sample in samples]
+            missing = [i for i, r in enumerate(routed) if r is None]
+            if missing:
+                raise ValueError(
+                    f"use_rollout_routing_replay: {len(missing)}/{len(samples)} samples missing "
+                    "rollout_routed_experts (see rollout logs for vLLM routing replay errors). "
+                    "Ensure vLLM 0.22+ serves with --enable-return-routed-experts."
+                )
+            train_data["rollout_routed_experts"] = routed
+        elif samples[0].rollout_routed_experts is not None:
```
aoshen02: Kinda weird, can we avoid passing `rollout_routed_experts` when `use_rollout_routing_replay` is false? I think it's inconsistent.

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -336,6 +338,10 @@ def _user_overrode(dest: str) -> bool:
         _, action = entry
         return getattr(args, dest, action.default) != action.default
 
+    # MoE routing replay (vLLM 0.22+, PR #39568): routed experts on ``/inference/v1/generate``.
```
aoshen02: I don't think we need to move the code

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -405,6 +411,83 @@ def _redact_cmd_for_log(cmd: list[str]) -> str:
     return " ".join(parts)
 
 
+def _response_json_or_fallback(response) -> dict:
+    """Parse JSON from an HTTP response, returning a structured error dict on failure."""
+    try:
+        data = response.json()
+    except (ValueError, json.JSONDecodeError):
+        return {"ok": False, "error": "Invalid JSON response", "raw": getattr(response, "text", "")}
```
aoshen02: Dead code

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -405,6 +411,83 @@ def _redact_cmd_for_log(cmd: list[str]) -> str:
...
+    if value is None:
+        return None
+    if isinstance(value, str):
+        return np.load(io.BytesIO(base64.b64decode(value)), allow_pickle=False)
+    if isinstance(value, list):
+        return np.asarray(value, dtype=np.int32)
+    return None
+
+
+def _verify_generate_routed_experts(base_url: str, model: str, timeout_s: float = 120.0) -> None:
```
aoshen02: I’m not sure `_verify_generate_routed_experts()` belongs in `VLLMEngine.init()`. I think we should move it to integration test

slime/rollout/vllm_rollout.py
```diff
@@ -491,7 +585,14 @@ async def generate(args: Namespace, sample: Sample, sampling_params: dict[str, A
         sample.rollout_log_probs = []
     sample.rollout_log_probs += new_response_log_probs
 
-    _apply_vllm_routed_experts(args, sample, output, choice)
+    _apply_vllm_routed_experts(
+        args,
+        sample,
+        output,
+        choice,
```
aoshen02: One issue with the current ordering: `_apply_vllm_routed_experts()` checks `sample.status == Sample.Status.ABORTED` to allow abort + 0-token samples without routed experts, but `sample.update_from_meta_info(args, meta)` is called only after `_apply_vllm_routed_experts()`.

So if vLLM returns `finish_reason="abort"` with no generated tokens/routed_experts, the sample is still `PENDING` when `_apply_vllm_routed_experts()` runs, and the intended abort guard will not trigger. This can raise `RuntimeError` during partial rollout abort/drain.

Could we either call `sample.update_from_meta_info(args, meta)` before `_apply_vllm_routed_experts()`, or make `_apply_vllm_routed_experts()` check the parsed finish reason directly instead of relying on `sample.status`?

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -1,10 +1,15 @@
...
 import ipaddress
+import json
 import logging
 import multiprocessing
 import os
 import time
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import numpy as np
```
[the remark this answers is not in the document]
aoshen02: make sense, could you remove this?

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -405,6 +411,83 @@ def _redact_cmd_for_log(cmd: list[str]) -> str:
...
+    if not isinstance(data, dict):
+        return {"ok": False, "error": "Response is not a dictionary", "data": data}
+    return data
+
+
+def _routing_rows_from_http_payload(value: Any) -> np.ndarray | None:
+    """Decode vLLM routed-experts HTTP field (base64 npy or nested list)."""
+    import base64
+    import io
+
```
aoshen02: Dead code

slime/rollout/vllm_rollout.py
```diff
@@ -161,40 +161,103 @@ def _vllm_meta_from_generate_choice(args: Namespace, choice: dict, usage: dict |
...
+    if isinstance(value, list):
+        return np.asarray(value, dtype=np.int32)
+    logger.warning("routed_experts payload must be nested list or base64 npy from vLLM HTTP API")
+    return None
+
 
-    vLLM ``/inference/v1/generate`` returns routed experts as a base64 encoded
-    ``.npy`` payload on each response choice when the server is launched with
-    ``--enable-return-routed-experts``.
+def _merge_generate_routed_experts(
```
aoshen02: Looking at vLLM’s `/inference/v1/generate` implementation, the response contract for routed experts seems much simpler than the compatibility logic here.

The protocol defines `choices[].routed_experts` as a single `str | None`, documented as base64-encoded `.npy` bytes with decoded shape `(num_tokens - 1, num_layers, num_experts_per_tok)`:
https://github.com/vllm-project/vllm/blob/71d810bbf44b34f3a019730a6878fbcbf2480499/vllm/entrypoints/serve/disagg/protocol.py#L157-L172

The serving code also always encodes it with `np.save(...)` + base64, and puts it only on `choice.routed_experts`:
https://github.com/vllm-project/vllm/blob/71d810bbf44b34f3a019730a6878fbcbf2480499/vllm/entrypoints/serve/disagg/serving.py#L277-L294

The vLLM test decodes it the same way:
https://github.com/vllm-project/vllm/blob/71d810bbf44b34f3a019730a6878fbcbf2480499/tests/entrypoints/serve/disagg/test_return_routed_experts.py#L70-L75

Given that contract, do we need to support nested-list payloads, `prompt_routed_experts`, or split/trim/concat logic in VIME? I think this path can be simplified to only read `choice["routed_experts"]`, decode the base64 `.npy`, validate `ndim == 3`, and require the row count to match `len(sample.tokens) - 1`.

slime/rollout/vllm_rollout.py
```diff
@@ -203,7 +266,36 @@ def _apply_vllm_routed_experts(
...
+    """Populate ``sample.rollout_routed_experts`` from ``/inference/v1/generate`` when R3 is enabled."""
+    if not getattr(args, "use_rollout_routing_replay", False):
+        return
+    arr = _merge_generate_routed_experts(output, choice, gen_token_count, prompt_token_count=prompt_token_count)
+    _assign_rollout_routed_experts(args, sample, arr)
+    if sample.rollout_routed_experts is not None:
+        return
+    if sample.status == Sample.Status.ABORTED and sample.response_length == 0:
+        return
+    pre = output.get("prompt_routed_experts")
```
aoshen02: vllm now doesn't have prompt_router experts so I think we can just simplify this logic.

slime/rollout/vllm_rollout.py
```diff
@@ -441,6 +533,7 @@ async def generate(args: Namespace, sample: Sample, sampling_params: dict[str, A
         gen_url = f"{base}/inference/v1/generate"
         with trace_span(sample, "vllm_mm_generate", attrs={"max_tokens": params["max_new_tokens"]}):
             output = await post(gen_url, generate_body, headers=headers)
+        request_prompt_len = len(generate_body.get("token_ids") or [])
```
aoshen02: vllm now doesn't have prompt_router experts so I think we can just simplify this logic.

slime/backends/vllm_utils/vllm_engine.py
CalvinXKY: > Dead code

Done

slime/ray/rollout.py
CalvinXKY: > Kinda weird, can we avoid passing `rollout_routed_experts` when `use_rollout_routing_replay` is false? I think it's inconsistent.

Good catch, removed

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -405,6 +411,83 @@ def _redact_cmd_for_log(cmd: list[str]) -> str:
...
+    import numpy as np
+
+    if value is None:
+        return None
+    if isinstance(value, str):
+        return np.load(io.BytesIO(base64.b64decode(value)), allow_pickle=False)
+    if isinstance(value, list):
+        return np.asarray(value, dtype=np.int32)
+    return None
```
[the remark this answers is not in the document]
CalvinXKY: Done

tests/unit/rollout/test_vllm_rollout.py
```diff
@@ -226,30 +226,72 @@ def test_decode_vllm_routed_experts_roundtrip():
...
 
 
-@pytest.mark.unit
-def test_apply_vllm_routed_experts_assigns_when_shape_matches():
-    arr = np.zeros((3, 2, 1), dtype=np.int32)
+def _encode_routed_npy(arr: np.ndarray) -> str:
     buf = io.BytesIO()
     np.save(buf, arr)
-    encoded = base64.b64encode(buf.getvalue()).decode("ascii")
+    return base64.b64encode(buf.getvalue()).decode("ascii")
```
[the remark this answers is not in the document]
CalvinXKY: Done

slime/backends/vllm_utils/vllm_engine.py
CalvinXKY: > I don't think we need to move the code

Agreed

slime/backends/vllm_utils/vllm_engine.py
CalvinXKY: > I’m not sure `_verify_generate_routed_experts()` belongs in `VLLMEngine.init()`. I think we should move it to integration test

That's a fair point — the main reason I kept it in init() is to fail fast at engine startup rather than discovering a broken routing configuration only when the first rollout batch arrives.

slime/rollout/vllm_rollout.py
CalvinXKY: > One issue with the current ordering: `_apply_vllm_routed_experts()` checks `sample.status == Sample.Status.ABORTED` to allow abort + 0-token samples without routed experts, but `sample.update_from_meta_info(args, meta)` is called only after `_apply_vllm_routed_experts()`.
> 
> So if vLLM returns `finish_reason="abort"` with no generated tokens/routed_experts, the sample is still `PENDING` when `_apply_vllm_routed_experts()` runs, and the intended abort guard will not trigger. This can raise `RuntimeError` during partial rollout abort/drain.
> 
> Could we either call `sample.update_from_meta_info(args, meta)` before `_apply_vllm_routed_experts()`, or make `_apply_vllm_routed_experts()` check the parsed finish reason directly instead of relying on `sample.status`?

 Fixed by calling sample.update_from_meta_info() before _apply_vllm_routed_experts(), so the abort guard based on sample.status works correctly. Also added a regression test.

slime/rollout/vllm_rollout.py
CalvinXKY: > vllm now doesn't have prompt_router experts so I think we can just simplify this logic.

Simplified to match vLLM's documented contract

slime/rollout/vllm_rollout.py
CalvinXKY: > vllm now doesn't have prompt_router experts so I think we can just simplify this logic.

Simplified to match vLLM's documented contract

slime/rollout/vllm_rollout.py
CalvinXKY: > Looking at vLLM’s `/inference/v1/generate` implementation, the response contract for routed experts seems much simpler than the compatibility logic here.
> 
> The protocol defines `choices[].routed_experts` as a single `str | None`, documented as base64-encoded `.npy` bytes with decoded shape `(num_tokens - 1, num_layers, num_experts_per_tok)`: https://github.com/vllm-project/vllm/blob/71d810bbf44b34f3a019730a6878fbcbf2480499/vllm/entrypoints/serve/disagg/protocol.py#L157-L172
> 
> The serving code also always encodes it with `np.save(...)` + base64, and puts it only on `choice.routed_experts`: https://github.com/vllm-project/vllm/blob/71d810bbf44b34f3a019730a6878fbcbf2480499/vllm/entrypoints/serve/disagg/serving.py#L277-L294
> 
> The vLLM test decodes it the same way: https://github.com/vllm-project/vllm/blob/71d810bbf44b34f3a019730a6878fbcbf2480499/tests/entrypoints/serve/disagg/test_return_routed_experts.py#L70-L75
> 
> Given that contract, do we need to support nested-list payloads, `prompt_routed_experts`, or split/trim/concat logic in VIME? I think this path can be simplified to only read `choice["routed_experts"]`, decode the base64 `.npy`, validate `ndim == 3`, and require the row count to match `len(sample.tokens) - 1`.

Simplified to match vLLM's documented contract

slime/backends/vllm_utils/vllm_engine.py
aoshen02: I think this belongs in a smoke/integration test rather than production init code.  rollout already validates missing `rollout_routed_experts` before training, so we still fail on the first real rollout anyway.

slime/rollout/vllm_rollout.py
```diff
@@ -161,49 +161,70 @@ def _vllm_meta_from_generate_choice(args: Namespace, choice: dict, usage: dict |
...
-        logger.warning(
-            f"routed_experts shape {arr.shape} does not match args (num_layers={nl}, moe_router_topk={mtk})",
-        )
-        return
-    sample.rollout_routed_experts = arr
+        raise RuntimeError(f"vLLM routing replay: routed_experts shape {arr.shape} != (rows,{nl},{mtk}) from args.")
+    sample.rollout_routed_experts = np.ascontiguousarray(arr.astype(np.int32, copy=True))
+
+
+def _vllm_expected_routed_rows_from_tokens(token_count: int) -> int:
```
aoshen02: dead code

tests/unit/backends/vllm_utils/test_vllm_engine.py
```diff
@@ -329,3 +304,36 @@ def test_init_weights_update_group_raises_after_three_failures(vllm_engine, monk
             group_name="g",
             backend="nccl",
         )
+
+
+@pytest.mark.unit
+def test_verify_generate_routed_experts_accepts_single_buffer(monkeypatch):
+    prompt_toks = 5
+    gen_toks = 3
```
aoshen02: We can remove the unit test and have an integration test about it.

slime/rollout/vllm_rollout.py
CalvinXKY: Done. Also cleaned up some dead code left over from debugging.

tests/unit/backends/vllm_utils/test_vllm_engine.py
CalvinXKY: OK.

CalvinXKY: After: https://github.com/vllm-project/vime/pull/49/changes/254056fbf44134cbe5a588643d0f61050e34f2e0

## E2E test works well：
image

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -559,7 +568,8 @@ def _init_normal(self) -> None:
             visible_devices=visible_devices,
             model_path=self.model_path,
         )
-        _wait_server_healthy(self._http_base(), process=self.process)
+        base = self._http_base()
+        _wait_server_healthy(base, process=self.process)
```
aoshen02: I think we can revert here.

CalvinXKY:
# CMP Test：

image

slime/backends/vllm_utils/vllm_engine.py
CalvinXKY: OK

PR #57, opened 2026-05-28
[Bugfix] [weight-sync] align colocated IPC coordinator/gather with vLLM engine slot

Meihan-chen:
## Summary

Related  https://github.com/vllm-project/vime/issues/56#issuecomment-4560065244

When the Megatron training topology does not match the colocated vLLM engine topology (e.g. `TP=4, CP=2, DP=1, EP=8` trainer + one 8-GPU vLLM engine), the colocated CUDA-IPC weight-sync path made two assumptions that only hold when Megatron TP spans the entire vLLM engine slot. Both are now fixed.

- **Coordinator selection** (`connect_rollout_engines`): the rank that issues `/start_weight_update` and `/finish_weight_update` was chosen by Megatron TP rank 0, which is not unique inside a vLLM engine slot when CP/DP is present (e.g. ranks `[0..7]` with `TP=4, CP=2` have TP ranks `[0,1,2,3,0,1,2,3]`). 
 This caused :
  ```text
  requests.exceptions.HTTPError: 500 Server Error for url: http://.../start_weight_update
  RuntimeError: start_weight_update called while a weight update is already active.
  Call finish_weight_update first.
  ```
    The coordinator is now the lowest **global** rank inside each engine GPU slot (`rank == start`), which is unique by construction.

- **IPC handle gather** (`_send_hf_chunk_via_ipc`): the all-gather of per-rank CUDA IPC payloads ran over Megatron's TP group, so the merged `update_info` only carried GPU UUIDs for half of the vLLM TP=8 workers, producing
  ```text
  ValueError: IPC handle not found for GPU UUID ...
   ```
   A dedicated Gloo process group is now created per engine slot covering the full `[start, end)` global rank range, and the gather + slot barrier run over that group. The merged payload now contains handles for every physical GPU UUID the colocated vLLM engine owns.

Single-GPU-slot configurations are unaffected (the existing `slot_size <= 1` fast path still uses `IPCWeightTransferEngine.trainer_send_weights` directly).

## Test plan

- [x] Unit tests under
      `tests/unit/backends/megatron_utils/update_weight/test_update_weight_from_tensor.py`
- [x] End-to-end smoke via the existing `tests/test_qwen3_30B_A3B.py` with `USE_DEEPEP=0`

### DIFF
--- a/slime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
+++ b/slime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
@@ -181,6 +181,8 @@ class UpdateWeightFromTensor:
         self._ipc_engine_coordinator: bool = False
         self._ipc_engine_slot_start: int | None = None
         self._ipc_engine_slot_end: int | None = None
+        self._ipc_engine_slot_group = None
+        self._ipc_engine_slot_ranks: list[int] | None = None
         self._distributed_engines: list[ActorHandle] = []
         self._model_update_groups = None
         self._is_distributed_src_rank: bool = False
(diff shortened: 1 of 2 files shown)


knlnguyen1802:
@aoshen02  I think this is already handle at https://github.com/vllm-project/vime/pull/48 right ?

aoshen02:
> @aoshen02 I think this is already handle at #48 right ?

Yah I think so, thus I will close it.

Issue #64, opened 2026-05-28
[Analysis] train_rollout_logprob_abs_diff: vLLM (vime) vs SGLang (slime) 


CalvinXKY:
### Summary

We ran a controlled comparison to understand **`train/train_rollout_logprob_abs_diff`** — the mean absolute difference between Megatron recomputed log-probs (`batch["log_probs"]` from `compute_log_prob`) and rollout-engine log-probs (`rollout_log_probs` from the inference backend) on the same tokens.

**Finding:** Under **aligned training hyperparameters** (same model, topology, batching, and GRPO settings), **vime + vLLM** keeps this metric **flat (~0.011)** across 30 rollout steps, while **upstream slime + SGLang** shows a **monotonic increase (~0.011 → ~0.036)**. Step-0 values match between stacks, which argues against a simple misconfiguration (dataset keys, batch size, temperature, etc.) and points to **train–inference alignment drift on the SGLang path** as training progresses. We plan to repeat on other prompt datasets; this issue captures the first Qwen3-4B run.

We would like to track analysis and possible mitigations (weight sync, CUDA graph / KV cache after `update_weights_from_distributed`, diagnostics) in vime/slime.

---

### Metric definition

From Megatron policy loss reporting (`slime/backends/megatron_utils/loss.py`):

```python
train_rollout_logprob_abs_diff = mean(|old_log_probs - rollout_log_probs|)
```

- `rollout_log_probs`: stored during rollout from the inference engine (vLLM or SGLang).
- `old_log_probs`: Megatron `compute_log_prob` on the **same rollout batch**, after weight sync to rollout engines and **before** the actor training step (not the training forward with gradients).

This metric is useful for detecting **Megatron ↔ rollout engine mismatch** (sync, numerics, logprob API semantics).

---

### Test design

| Item | vime side | slime side |
|------|-----------|------------|
| Model | **Qwen3-4B** (Megatron: 36 layers, hidden 2560, GQA, `rotary-base` 1e6, etc.) | Same |
| Prompt data | Parquet + `--apply-chat-template`, `question` / `label`, `--rollout-shuffle`, `--rm-type deepscaler` | Same |
| Topology | **4 GPUs train + 4 GPUs rollout** (no `--colocate`) | Same |
| Rollout engines | **2× TP=2** on 4 rollout GPUs (`--rollout-num-gpus 4`, `--rollout-num-gpus-per-engine 2`) | Same |
[...]

PR #69, opened 2026-05-29
[Bugfix] Fix high train_rollout_logprob_abs_diff

andakai:
Try to fix the issue here. The details are also in this issue. https://github.com/vllm-project/vime/issues/64

### DIFF
--- a/slime/backends/vllm_utils/vllm_engine.py
+++ b/slime/backends/vllm_utils/vllm_engine.py
@@ -345,7 +345,12 @@ def launch_server_process(
         gpu_mem = 0.55  # vime preferred
     cmd += ["--gpu-memory-utilization", str(gpu_mem)]
 
-    # 2) weight_transfer_config: vllm default None disables /init_weight_transfer_engine,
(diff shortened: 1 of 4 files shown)


knlnguyen1802:
@andakai  PTAL at bot comment and fix DCO

aoshen02:
`top_k=-1` is accepted by vLLM as "disable top-k", and forwarding it is consistent with the user-facing rollout config. Please update the existing unit test that still expects `top_k=-1` to be omitted;
     the new expected behavior should assert that `top_k=-1` is present in the vLLM sampling params.

aoshen02:
Defaulting to `processed_logprobs` makes sense for the default rollout setup because vLLM raw logprobs are pre-temperature while Megatron replay scales logits by rollout temperature. However, this is
     only semantically aligned when top-k/top-p filtering is disabled (`top_p=1`, `top_k=-1/0`). With `top_p < 1` or `top_k > 0`, vLLM processed logprobs are computed after filtering/renormalization, while
     Megatron replay appears to only apply temperature. We can emit a warning when `processed_logprobs` is used together with non-default top-p/top-k.

PR #74, opened 2026-05-29
examples: remove retool and search-r1 examples

aoshen02:
## What

Removes two end-to-end examples and their `examples/README.md` index entries:

- **`examples/retool/`** — tool-calling RL+SFT example (custom rollout via `generate_with_retool.py` + code sandbox).
- **`examples/search-r1/`** — retrieval-augmented multi-turn example (custom rollout via `generate_with_search.py` + dense/google retrieval servers).

18 files, ~3067 lines deleted.

## Why

Trimming unmaintained end-to-end examples. Neither depends on the (now-removed) SGLang runtime — both ride on `--rollout-function-path` with their own generate functions — so this is an examples-trim, not part of the SGLang cleanup.

## Dangling-reference check

On current `main`, the `docs/{en,zh}/index.rst` toctree entries and the `customization.md` link into `examples/search-r1/` are already gone upstream. After this PR the **only** remaining mentions are in the "Multiturn Adaptation" section of `docs/{en,zh}/get_started/quick_start.md`, which cite the **external** `github.com/PeterGriffinJin/Search-R1` repo — not the deleted in-repo dir. So **no in-repo links are left dangling.**

## Open question for reviewers

The "Multiturn Adaptation" section in `quick_start.md` is a generic multi-turn-adaptation tutorial that *uses* Search-R1 as its worked example but does not link the deleted code. Left intact here. If we'd rather not reference a no-longer-shipped example, that section can be reworded/removed in a follow-up — flagging rather than gutting a conceptual doc section in this PR.


CalvinXKY:
LGTM

PR #75, opened 2026-05-30
[Example][Bugfix]: align custom rollout paths with vLLM router API

CalvinXKY:
## Summary

Align remaining vime examples and related rollout hooks with the vLLM-first API surface after the SGLang removal on `main`. Custom generate paths now use `args.router_ip` / `args.router_port` instead of the deprecated `vllm_router_*` attributes, fixing smoke failures in examples such as `multi_agent`.

## Changes

- **Examples**: update `multi_agent/agent_system.py` and `geo3k_vlm_multi_turn/rollout.py` to call the router via `router_ip` / `router_port`
- **Docs**: fix stale router field names in usage guides
- **Framework (minor)**: align `--opd-type vllm` with example scripts and teacher scoring via `/v1/completions` + `prompt_logprobs`

## Validation

Smoke-tested on **A100-server / vime_v22** using external scripts under `run_script/validate_examples/` (not included in this PR):

| Example | Result |
|---------|--------|
| train_infer_mismatch_helper | PASS |
| eval_multi_task | PASS |
| fully_async | PASS |
| multi_agent | PASS (after router fix) |
| geo3k_vlm | PASS |
| on_policy_distillation (megatron smoke) | PASS |
| geo3k_vlm_multi_turn | FAIL (example-level multimodal processor issue) |

Removed on `main` and not covered here: `retool`, `search-r1` (SGLang-only).

## Test plan

- [ ] `multi_agent` example completes at least one rollout
- [ ] `geo3k_vlm_multi_turn` and `geo3k_vlm` smoke on available VLM checkpoints
- [ ] No regressions in existing CI e2e tests (this PR does not modify CI test files)

### DIFF
--- a/examples/geo3k_vlm_multi_turn/rollout.py
+++ b/examples/geo3k_vlm_multi_turn/rollout.py
@@ -8,7 +8,7 @@ from typing import Any
 
 import torch
 
-# When executed as a module: python -m examples.vlm_multi_turn.rollout
+from examples.geo3k_vlm_multi_turn.base_env import BaseInteractionEnv
 from vime.rollout.vllm_rollout import (
     GenerateState,
     _build_inference_sampling_params,
@@ -16,10 +16,16 @@ from vime.rollout.vllm_rollout import (
     _mm_render_response_to_generate_body,
 )
 from vime.utils.http_utils import post
-from vime.utils.processing_utils import encode_image_for_rollout_engine
+from vime.utils.processing_utils import build_processor_kwargs, encode_image_for_rollout_engine
 from vime.utils.types import Sample
 
 DEFAULT_ENV_MODULE = "examples.vlm_multi_turn.env_geo3k"
+
+# Dummy messages used for calculating trim length in chat template encoding.
+DUMMY_MESSAGES = [
+    {"role": "system", "content": "You are a helpful assistant."},
+    {"role": "user", "content": "I am a user."},
+]
 
 
 def _load_env_module(env_path: str | None):
@@ -62,17 +68,18 @@ def _content_to_render_format(content: list[dict]) -> list[dict]:
 
 
 def _build_initial_user_message(sample: Sample) -> dict:
-    """Build the initial user message from sample.prompt + sample.multimodal_inputs.images."""
-    content: list[dict] = []
+    """Build the initial user message for the render route (same layout as vllm_rollout MM render)."""
     images = (sample.multimodal_inputs or {}).get("images") or []
+    if not images:
+        return {"role": "user", "content": sample.prompt}
+    content: list[dict] = [{"type": "text", "text": sample.prompt}]
     for image in images:
         content.append({"type": "image", "image": image})
-    content.append({"type": "text", "text": sample.prompt})
     return {"role": "user", "content": content}
 
 
 def _messages_for_render(messages: list[dict]) -> list[dict]:
-    """Normalize per-turn messages to the render-route shape (image → image_url)."""
+    """Normalize per-turn messages to the render-route shape (image -> image_url)."""
     out: list[dict] = []
     for msg in messages:
         content = msg.get("content")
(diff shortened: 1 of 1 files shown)


CalvinXKY:
## A100 smoke validation (vime_v22 @ A100-server)

Re-ran `run_script/validate_examples/run_all.sh` on **2026-05-30** against `feature/examples-validation` (PR branch files synced to NFS).

| Example | Result |
|---------|--------|
| train_infer_mismatch_helper | PASS |
| eval_multi_task | PASS |
| fully_async | PASS |
| multi_agent | PASS |
| on_policy_distillation_megatron | PASS |
| geo3k_vlm | PASS |
| geo3k_vlm_multi_turn | PASS (after fix in latest commit) |

**Log**: `/data/nfs_87/xky/new_rl/logs/examples_validate/run_all_20260530_102300.log`

### geo3k_vlm_multi_turn fix

Previous failure was `IndexError` in Qwen3VL processor (`image_grid_thw` mismatch): the rollout re-applied `apply_chat_template` on a message that already contained separate image entries plus a chat-templated `sample.prompt`, duplicating vision slots.

Fixed by aligning with `vllm_rollout`:
- processor path: `processor(text=sample.prompt, **build_processor_kwargs(multimodal_inputs))`
- render message layout: text first, then image entries (same as standard MM render)

Retest log: `smoke_geo3k_multi_turn_retest_20260530.log` → **PASS**

### Out of scope (unchanged)

| Example | Status |
|---------|--------|
| search_r1 | SKIP (missing Search-R1 data) |
| strands_sglang / tau_bench | SKIP (deps not in vime_v22) |
| retool / search-r1 | N/A (removed on main, SGLang-only) |

aoshen02:
I checked the corresponding `geo3k_vlm_multi_turn` code in upstream slime and miles, and I think this PR is fixing the right root cause for the initial prompt path: `sample.prompt` is already the dataset-rendered / chat-templated prompt, so the first turn should not wrap it back into a fresh chat message and run `apply_chat_template` again.

References:

- slime keeps the initial prompt path as `processor(text=sample.prompt, ...)`, not `apply_chat_template([message])`: https://github.com/THUDM/slime/blob/bf14dc21/examples/geo3k_vlm_multi_turn/rollout.py#L154-L167
- miles has the same shape: https://github.com/radixark/miles/blob/b40564d71/examples/geo3k_vlm_multi_turn/rollout.py#L154-L167

So the change to build initial train-side features from `sample.prompt` + `sample.multimodal_inputs` is the right direction.

One thing I would still adjust before merging: please keep the multi-turn behavior closer to slime/miles instead of dropping it entirely.

Concrete suggestions:

1. Keep separate helpers for the two cases:
   - initial dataset prompt: use `processor(text=sample.prompt, **build_processor_kwargs(sample.multimodal_inputs))`
   - later env observations: they are freshly-created messages from `env.format_observation(...)`, so it is still reasonable to process those messages separately, as slime/miles do.

2. Preserve observation multimodal features. This PR removes the old `obs_train_feats` collection. That avoids the initial double-template bug, but it also means later observation images/videos will no longer contribute to `sample.multimodal_train_inputs`. In slime/miles, the observation path still returns and merges per-turn `multimodal_train_inputs`: https://github.com/THUDM/slime/blob/bf14dc21/examples/geo3k_vlm_multi_turn/rollout.py#L212-L233

3. Preserve the full token trajectory if possible. The slime/miles implementation appends prompt/observation tokens with `loss_mask=0` and assistant tokens with `loss_mask=1`. In the current vLLM render path, `sample.tokens` only grows by generated assistant tokens, while the prompt and observation turns are only implicit in the render request. A closer mirror would be:
   - after `/v1/chat/completions/render`, keep the rendered prompt `token_ids`
   - append only the delta not already in `sample.tokens` with `loss_mask=0` / logprob `0.0`
   - append generated tokens with `loss_mask=1`
   - validate prefix alignment so silent template drift is caught early

4. Add a regression test for the exact failure mode:
   - initial `sample.prompt` is already chat-templated and has multimodal inputs
   - the custom rollout must not call `apply_chat_template([initial_message])` for processor features
   - later observation messages can still be processed as fresh messages
   - the number/order of image placeholders matches the processor output, especially `image_grid_thw`

In short: I would keep this PR's initial-prompt fix, but mirror slime/miles by preserving the observation multimodal path and full multi-turn token/loss-mask trajectory. That gives us the bug fix without losing the behavior the original slime/miles rollout was maintaining.

CalvinXKY:
Related to  #120.

PR #96, opened 2026-05-31
perf: chunk Megatron TP-side grad all-reduce to bound peak memory

aoshen02:
## What

Backports slime **#1899**.

Patches `_allreduce_non_tensor_model_parallel_grads` (and its legacy alias `_allreduce_layernorm_grads`) in `megatron.core.distributed.finalize_model_grads` to coalesce/all-reduce TP-side gradients in **size-bounded chunks** (`SLIME_GRAD_COALESCE_CHUNK_BYTES`, default 1 GiB) instead of a single large `_flatten_dense_tensors(grads)`.

## Why

The default path flattens *all* TP-side grads into one contiguous buffer before the all-reduce. For large models that buffer can be huge, and under allocator fragmentation the contiguous allocation OOMs even when total free memory is sufficient. Chunking bounds the peak contiguous allocation.

SUM/AVG reductions are element-wise, so chunking is **mathematically equivalent** — no numerical change, purely a memory-pressure mitigation.

## Shape / safety

- Self-contained: new `slime/backends/megatron_utils/megatron_patch/` subpackage, wired via a single `from . import megatron_patch` in `megatron_utils/__init__.py`.
- The patch module guards its megatron imports in `try/except`; if megatron is unavailable or its symbol layout changed, it logs a warning and is a **no-op** (never crashes import).
- Cross-compatible across the `core_v0.13.0` and post-`core_v0.15.0rc7` Megatron lines — API differences (`use_custom_fsdp` vs `use_megatron_fsdp`, `_get_main_grad_attr` arity, target-fn signature) are resolved at runtime, no version-conditional imports.

Patch file is byte-identical to slime upstream.

## Scope note

This is **not** a precision change — it's OOM/numerical-safety infra (the P2 item from the upstream survey). Opened separately per request.

## Decoupling

Independent of #92 / #93 / #94 / #95 — new subpackage + one import line; only shared file is `__init__.py` which none of the others touch.

### DIFF
--- a/slime/backends/megatron_utils/__init__.py
+++ b/slime/backends/megatron_utils/__init__.py
@@ -40,3 +40,5 @@ except ImportError:
     pass
 
 logging.getLogger("megatron").setLevel(logging.WARNING)
+
+from . import megatron_patch  # noqa: F401, E402
--- a/slime/backends/megatron_utils/megatron_patch/__init__.py
+++ b/slime/backends/megatron_utils/megatron_patch/__init__.py
@@ -1,0 +1,1 @@
+from . import megatron_chunked_grad_coalesce_patch  # noqa: F401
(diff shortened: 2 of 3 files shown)


CalvinXKY:
LGTM

PR #99, opened 2026-05-31
Fix OPD teacher endpoint for vLLM test

aoshen02:


aoshen02:
Superseded by #141. Root cause is the built-in OPD reward_func sending vime's disaggregated /inference/v1/generate payload, while the test launches a stock vllm openai api_server that only serves /v1/completions (the contract documented by merged #62). This PR flipped the test's --rm-url to /inference/v1/generate, but the stock teacher doesn't serve that route (would 404). #141 instead fixes the *code* to speak /v1/completions + echo + prompt_logprobs=1 and read choices[0].prompt_logprobs; validated e2e on gb200 (EXIT_RC=0, teacher_log_probs flow into opd_reverse_kl). Closing in favor of #141.

aoshen02:
Correction to my earlier close comment: I was wrong that flipping the URL to /inference/v1/generate would 404. That route IS mounted on the stock openai api_server in this vendored vLLM (build_app attaches the disagg router for any generate-capable model), so #99's approach was actually viable. The real reason #141 prefers /v1/completions is portability (it exists on any vLLM incl. upstream and any OpenAI-compatible server; /inference/v1/generate is a fork-only endpoint under entrypoints/serve/disagg/) plus matching the contract already documented by merged #62. The 400 itself came from the test pointing --rm-url at /v1/completions while the code POSTed an /inference/v1/generate-shaped body. Still closing in favor of #141 (validated e2e), just correcting the record.

Issue #11, opened 2026-05-18
[RFC] VIME Roadmap


gcanlin:
# [Roadmap]

###  Port remaining content from slime (fork point `8ef1fb47` → snapshot `44d29ee`) that hasn't landed on vime main.

- [ ] **Delta weight sync — [#278](https://github.com/vllm-project/vime/pull/278)**
  Bandwidth-optimized non-colocate weight transfer from slime #1806/#1946/#1991. Bytewise-diff encoding (indices / gap-deltas / zstd), sparse NCCL broadcast or disk-safetensors transport. Includes:
  - `update_weight_from_distributed_delta.py` — core delta encoder/decoder + NCCL publisher/subscriber
  - `delta_io.py` — standalone `DeltaEncoding`/`DeltaParam`/`DeltaSpec` dataclasses
  - `actor.py` / `data.py` / `train_metric_utils.py` — delta mode branch + `extra_metrics` plumbing to wandb
  - `arguments.py` — 7 CLI args (`--update-weight-mode`, `--update-weight-transport`, `--update-weight-encoding`, etc.)
  - `types.py` — `rollout_id` field on `Sample`
  - docs (en/zh), example script, unit test

- [ ] **Run scripts — [#260](https://github.com/vllm-project/vime/pull/260)**
  Complete slime-exact translation of all 29 training scripts following `SGLANG_TO_VLLM_TRANSLATION.md`. 23 new scripts + 6 existing updated. Covers dense, MoE, FP8, SFT, multi-node, and speculative decoding configurations.

- [ ] **Other Missing examples** 
  Slime examples not yet on vime main (need new PR):
  - `eval_multi_task/` — multi-task evaluation pipeline
  - `on_policy_distillation/` — OPD example scripts
  - `retool/` — ReTool tool-call RL example
  - `search-r1/` — retrieval-augmented rollout example

### Disaggregation (PD / EPD / OPD)

- [ ] **EPD (encoder prefill disaggregation)** `P2` — VLM encoder server split from decoder prefill; composes with PD. (Net-new in vime; no slime counterpart.)
  Acceptance: new vLLM-only test gated on multimodal landing.
- [ ] Residual rollout overhead profiling (py-spy RolloutManager) — the remaining ~3-4% gap after router + logprobs fixes is in framework overhead, not engine/router

- [ ] LoRA support — [#228](https://github.com/vllm-project/vime/pull/228)

## Runtime State Switching
[...]

PR #104, opened 2026-06-01
[feat] Implement update weight chunk for vime, follow Sky-RL

knlnguyen1802:
Part of https://github.com/vllm-project/vime/issues/11

## Summary

Refactors `VLLMEngine.update_weights_from_tensor` and `VLLMEngine.update_weights_chunk` in
[`slime/backends/vllm_utils/vllm_engine.py`](slime/backends/vllm_utils/vllm_engine.py) to
route their `collective_rpc → update_weights_chunk` POSTs through the shared
`_make_request` helper, instead of hand-rolling `_post_json` + `raise_for_status()` +
`response.json()` at each call site.

This aligns slime's vLLM weight-transfer control plane with the pattern used in SkyRL's
[`RemoteInferenceClient.update_weights_chunk`](https://github.com/NovaSky-AI/SkyRL/blob/36f38c745c117c7b66fd5079038dbd5d8193e802/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py#L1079):
a single helper owns request issuance, HTTP-error surfacing, empty-body handling, and
JSON parsing — callers only describe the RPC.

## Motivation

- **Single choke point.** `_make_request` is already the documented control-plane POST
  helper (used by `start_weight_update`, `finish_weight_update`,
  `init_weight_transfer_engine`, `update_weights_from_distributed`, etc.). The two
  chunk paths were the last stragglers still calling `_post_json` directly.
- **Consistent error reporting.** `_make_request` → `_response_json` attaches
  `response.text` to `HTTPError` via `e.add_note(...)`, so a 4xx/5xx from vLLM now
  surfaces the server message instead of a bare status line.
- **Correct empty-body handling.** vLLM occasionally returns `200` with an empty body
  on `collective_rpc`. `_response_json` normalises that to `{"ok": True}`; the old
  inline `response.json()` raised `JSONDecodeError` and was masked into
  `{"ok": True, "raw": response.text}` — losing structured info on real successes.
- **Less code.** −12 / +4 lines, no behaviour change on the happy path.
- **Follows SkyRL.** Brings the API surface of `update_weights_chunk` in line with
  SkyRL's `RemoteInferenceClient`, easing future cross-project porting.

## Changes

- `update_weights_from_tensor`: now calls
  `self._make_request("collective_rpc", {...}, timeout=...)` and returns the parsed
  dict directly. `_weight_version` is still recorded only after the call returns (so a
  raised exception cannot advance the version — preserved invariant).
- `update_weights_chunk`: same migration; headless-worker (`node_rank > 0`)
  short-circuit and the `ipc_handles → ipc_handles_pickled` cloudpickle+base64
  serialisation are unchanged.

### Diff (high level)

```diff
- response = self._post_json("collective_rpc", {...}, timeout=update_timeout_s)
- response.raise_for_status()
- try:
-     return response.json()
- except Exception:
-     return {"ok": True, "raw": response.text}
+ response = self._make_request("collective_rpc", {...}, timeout=update_timeout_s)
+ return response
```

## Backwards compatibility

- Wire format on `POST /collective_rpc` is **identical**
  (`{"method": "update_weights_chunk", "kwargs": {"update_info": ...}}`).
- Return type is still `dict`. The only observable difference is that an empty-body
  `200` now returns `{"ok": True}` instead of `{"ok": True, "raw": ""}`.
- `SLIME_VLLM_WEIGHT_TRANSFER_HTTP_TIMEOUT_SEC` env override is preserved.

## Risk

Low. Control-plane only; no rollout / training-loop code touched. The refactor is a
pure delegation to an existing, already-exercised helper.

---

### Unit tests (In progress)

Add to `tests/unit/backends/vllm_utils/test_vllm_engine.py`:

1. `test_update_weights_from_tensor_uses_make_request_collective_rpc` — asserts the
   new path: endpoint == `collective_rpc`, payload wraps
   `{"method": "update_weights_chunk", "kwargs": {"update_info": ...}}`, `ipc_handles`
   is cloudpickled into `ipc_handles_pickled`, the function returns `_make_request`'s
   dict verbatim, and `_weight_version` is recorded.
2. `test_update_weights_from_tensor_does_not_advance_version_on_make_request_failure`
   — guards the resync-safety invariant: on `_make_request` raising, `_weight_version`
   must remain at its old value.
3. `test_update_weights_chunk_uses_make_request_and_returns_dict` — mirrors (1) for
   the standalone `update_weights_chunk` entry point.
4. `test_update_weights_chunk_noop_on_headless_worker` — `node_rank > 0` workers still
   short-circuit to `{"ok": True, "skipped": True}` with **zero** HTTP traffic.

### Integration test (required before merge)

End-to-end weight-transfer correctness must be validated on a real vLLM rollout
engine, because the unit tests mock `_make_request`. **Target model:
`Qwen/Qwen3-4B`** — the smallest model in the repo that exercises the chunked
tensor-IPC update path on 8 GPUs.

Run the existing 4B PPO smoke test, which performs multiple
`update_weights_from_tensor → collective_rpc/update_weights_chunk` round-trips per
rollout step:

```powershell
# Single-node, 8×GPU, exercises start_weight_update → update_weights_chunk (xN) → finish_weight_update
python tests/test_qwen3_4B_ppo.py
```

And, to cover the save / load + async-save flow that also drives chunked updates:

```powershell
python tests/test_qwen3_4B_ckpt.py
python tests/test_qwen3_4B_ckpt.py --async-save
```

## Checklist

- [ ] Unit tests added and passing locally.
- [ ] `tests/test_qwen3_4B_ppo.py` passes on an 8×GPU node.
- [ ] `tests/test_qwen3_4B_ckpt.py` (sync + `--async-save`) passes.
- [ ] No new lint / type errors in `slime/backends/vllm_utils/vllm_engine.py`.

Performance comparison

Test on `Qwen3-0.4B` with TP=2, 4 A100 cards

Current code
elapsed_trend_distribution

Avg update time after 200 steps: 4.01s 

New code
elapsed_trend_distribution_new

Avg update time after 200 steps: 3.71s


vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -461,3 +461,85 @@ class vLLMColocateWorkerExtension:
     def __new__(cls, **kwargs):
         _VLLMHijack.hijack()
         return super().__new__(cls)
+
+    # ── Three-phase weight update protocol ────────────────────────────────────
+    # Mirrors SkyRL's NewInferenceWorkerWrap. Callable via /collective_rpc from
+    # VLLMEngine.update_weights / update_weights_chunk on the trainer side.
+
+    def update_weights_chunk(self, update_info: dict) -> None:
```
CalvinXKY, line 469: Could we use POST /collective_rpc with method="update_weights" instead of method="update_weights_chunk"? The native /update_weights endpoint already fans out to collective_rpc("update_weights") → gpu_worker.update_weights() → IPCWeightTransferEngine.receive_weights() (with {uuid: ipc_args} + rebuild_cuda_tensor). The update_weights_chunk path duplicates that logic in the worker extension and switches the handle format to (func, args), which is incompatible with receive_weights. Is there a specific reason we can't stay on the official path?

knlnguyen1802: It's possible but I think vllm might support it too. Currently sky-RL doing like this but we can make it simpler here

slime/backends/vllm_utils/vllm_engine.py
```diff
@@ -825,11 +825,53 @@ def update_weights_from_tensor(
...
+        import base64
+
+        import cloudpickle
+
+        payload = dict(update_info)
+        if payload.get("ipc_handles") is not None:
+            payload["ipc_handles_pickled"] = base64.b64encode(
+                cloudpickle.dumps(payload.pop("ipc_handles"))
+            ).decode("utf-8")
+        update_timeout_s = float(os.environ.get("SLIME_VLLM_WEIGHT_TRANSFER_HTTP_TIMEOUT_SEC", "900"))
```
CalvinXKY: We hit failures due to a bug where vLLM returns HTTP 200 with an empty body. main already has a fix — after rebasing you should see _weight_transfer_http_timeout. The SLIME_VLLM_WEIGHT_TRANSFER_HTTP_TIMEOUT_SEC has been removed

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
knlnguyen1802: I think this is design for skyRL and it's better than current method in vllm. Let's me see if there is better way to adapt this instead of copy it all from skyRL

CalvinXKY: OK

CalvinXKY (approved):
LGTM.   

Checked:

 ## Environment

- **Server**: A800-server2 (7.216.196.62)
- **Container**: vime_debug (8×A800-80GB)
- **Model**: Qwen3-4B
- **Config**: TP=2, CP=2, colocate, --ci-test

params:
```
export PYTHONUNBUFFERED=1

VIME_ROOT="/data/nfs_87/xky/pr104_check/vime_pr104"
cd "${VIME_ROOT}"

source "${VIME_ROOT}/scripts/models/qwen3-4B.sh"

HF_CKPT="/data/nfs_87/xky/models/Qwen3-4B"
REF_LOAD="/data/nfs_87/xky/models/Qwen3-4B_torch_dist"
SAVE_DIR="/data/nfs_87/xky/pr104_test/checkpoints/qwen3_4b_colocate_pr104"

LOG_ROOT="/data/nfs_87/xky/pr104_test/logs"
TS="$(date +%Y%m%d_%H%M%S)"
export TENSORBOARD_DIR="${LOG_ROOT}/tb_pr104_${TS}"
LOG_FILE="${LOG_ROOT}/train_pr104_${TS}.log"
mkdir -p "${TENSORBOARD_DIR}" "${LOG_ROOT}" "${SAVE_DIR}"

CKPT_ARGS=(
   --hf-checkpoint "${HF_CKPT}"
   --ref-load "${REF_LOAD}"
   --save "${SAVE_DIR}"
   --save-interval 100
)

ROLLOUT_ARGS=(
   --prompt-data /root/datasets/dapo-math-17k/dapo-math-17k.jsonl
   --input-key prompt
   --label-key label
   --apply-chat-template
   --rollout-shuffle
   --rm-type deepscaler
   --num-rollout 3
   --rollout-batch-size 8
   --n-samples-per-prompt 4
   --rollout-max-response-len 2048
   --rollout-temperature 0.8
   --global-batch-size 16
   --balance-data
)

PERF_ARGS=(
   --sequence-parallel
   --context-parallel-size 2
   --expert-model-parallel-size 1
   --expert-tensor-parallel-size 1
   --recompute-granularity full
   --recompute-method uniform
   --recompute-num-layers 1
   --use-dynamic-batch-size
   --max-tokens-per-gpu 4096
)

PPO_ARGS=(
   --advantage-estimator ppo
   --kl-loss-coef 0.00
   --kl-loss-type k1
   --kl-coef 0.00
   --entropy-coef 0.00
   --eps-clip 4e-4
   --num-critic-only-steps 1
   --normalize-advantages
)

OPTIMIZER_ARGS=(
   --optimizer adam
   --lr 1e-6
   --lr-decay-style constant
   --weight-decay 0.1
   --adam-beta1 0.9
   --adam-beta2 0.98
)

VLLM_ARGS=(
   --rollout-num-gpus-per-engine 2
   --rollout-num-gpus 8
   --vllm-gpu-memory-utilization 0.7
   --vllm-max-num-seqs 512
   --vllm-max-cudagraph-capture-size 32
)

CI_ARGS=(
   --ci-test
)

MISC_ARGS=(
   --attention-dropout 0.0
   --hidden-dropout 0.0
   --accumulate-allreduce-grads-in-fp32
   --attention-softmax-in-fp32
   --attention-backend flash
   --train-memory-margin-bytes 2147483648
   --actor-num-nodes 1
   --actor-num-gpus-per-node 8
   --colocate
)

NUM_GPUS=8

export MASTER_ADDR="${MASTER_ADDR:-127.0.0.1}"
ray start --head \
  --node-ip-address "${MASTER_ADDR}" \
  --num-gpus "${NUM_GPUS}" \
  --disable-usage-stats \
  --dashboard-host=0.0.0.0 \
  --dashboard-port=8265

NVLINK_COUNT=$(nvidia-smi topo -m 2>/dev/null | grep -o 'NV[0-9][0-9]*' | wc -l)
HAS_NVLINK=0
if [ "$NVLINK_COUNT" -gt 0 ]; then
  HAS_NVLINK=1
fi

RUNTIME_ENV_JSON="{
  \"env_vars\": {
    \"PYTHONPATH\": \"${VIME_ROOT}:/root/Megatron-LM/\",
    \"CUDA_DEVICE_MAX_CONNECTIONS\": \"1\",
    \"NCCL_NVLS_ENABLE\": \"${HAS_NVLINK}\"
  }
}"

ray job submit --working-dir /data/nfs_87/xky/pr104_check/vime_pr104 --address="http://127.0.0.1:8265" \
   --runtime-env-json="${RUNTIME_ENV_JSON}" \
   -- python3 train.py \
   --train-backend megatron \
   ${MODEL_ARGS[@]} \
   ${CKPT_ARGS[@]} \
   ${ROLLOUT_ARGS[@]} \
   ${OPTIMIZER_ARGS[@]} \
   ${PPO_ARGS[@]} \
   ${PERF_ARGS[@]} \
   ${VLLM_ARGS[@]} \
   ${CI_ARGS[@]} \
   ${MISC_ARGS[@]} \
   2>&1 | tee -a "${LOG_FILE}"

```

##  Results

| Check | Status | Evidence |
|-------|--------|----------|
| `vLLMColocateWorkerExtension` injected with `update_weights_chunk` | PASS | Log: "Injected <class '...vLLMColocateWorkerExtension'> into <class '...Worker'> for extended collective_rpc calls ['update_weights_chunk']" |
| `start_weight_update` HTTP 200 | PASS | Multiple `POST /start_weight_update HTTP/1.1" 200 OK` entries |
| `collective_rpc` (update_weights_chunk) HTTP 200 | PASS | 63× `POST /collective_rpc HTTP/1.1" 200 OK` per engine per cycle (36 layers × ~2 shards) |
| `finish_weight_update` HTTP 200 | PASS | Multiple `POST /finish_weight_update HTTP/1.1" 200 OK` entries |
| 3+ complete weight update cycles | PASS | At least 3 full `start→chunk→finish` cycles observed |
| Training steps completed | PASS | Steps 1-5 completed with healthy metrics |
| `update_weights_time` reasonable | PASS | 1.67-3.35s per cycle |
| No errors/crashes during weight sync | PASS | No RuntimeError, HTTP errors, or CUDA errors in logs |

Image

## Training Metrics (Sample)

```
step 2: train/loss=-0.158, train/pg_loss=-0.158, train/grad_norm=1.174
step 3: train/loss=0.132, train/pg_loss=0.132, train/grad_norm=0.890
step 5: train/loss=0.205, train/pg_loss=0.205, train/grad_norm=0.906

perf 1: update_weights_time=3.35s, step_time=115.6s
perf 2: update_weights_time=1.67s, step_time=65.8s
```

Performance unchanged — the hot paths (serialization, IPC reconstruction, weight loading) are identical. Two non-blocking notes: (1) PR description understates scope (84 lines of new worker logic + reduce_tensor contract change, not just a refactor), (2) _post_vllm_update_weights_http is retained but not mentioned.

PR #21, opened 2026-05-21
Sync to vLLM 0.21.0

CalvinXKY:
## Purpose

This PR adapts vime’s Megatron→vLLM **native** weight sync to **vLLM 0.21.x**, as part of the image / dependency upgrade tracked in [Issue #20](https://github.com/vllm-project/vime/issues/20) (vllm 0.21 + torch 2.11 stack).

vLLM 0.21 exposes a four-phase weight transfer control plane ([vLLM #39212](https://github.com/vllm-project/vllm/pull/39212)):

- `POST /start_weight_update`
- `POST /update_weights`
- `POST /finish_weight_update`

Trainer-side sends use `NCCLTrainerSendWeightsArgs` with `NCCLWeightTransferEngine.trainer_send_weights(...)`.

In-process NCCL on the trainer (replacing the subprocess bridge) is already covered by [#13](https://github.com/vllm-project/vime/pull/13); this PR builds on that and focuses on **0.21 HTTP/API alignment** plus **unit test coverage**.

### vLLM 0.21 weight sync (`update_weight_from_distributed.py`)

- Wrap each full sync with rank-0 `start_weight_update(is_checkpoint_format=True)` → per-bucket `/update_weights` + `trainer_send_weights(NCCLTrainerSendWeightsArgs(...))` → `finish_weight_update`, with Gloo barriers.
- Per-chunk `update_info` no longer carries legacy `is_checkpoint_format` (handled in `start`).

### `VLLMEngine` (`vllm_engine.py`)

- Always call vLLM 0.21 start/finish/update routes (no legacy protocol branch).
- Shared `_response_json_or_fallback()`, unified weight-transfer HTTP timeout, `_http_base()` init guard.

### Unit tests (new layout)

Introduce layered unit tests under `tests/unit/` (pytest `unit` marker unchanged):

| Path | Scope |
|------|--------|
| `tests/unit/backends/vllm_utils/test_vllm_engine.py` | vLLM 0.21 HTTP four-phase endpoints, helpers (`_normalize_vllm_wake_tags`, `_serialize_for_cli`, init retry, IPv6 base URL, etc.) — **28 cases**, no live vLLM server |
| `tests/unit/backends/vllm_utils/test_arguments.py` | Moved from `tests/test_vllm_arguments.py` (CLI / router args) |
[...]

PR #110, opened 2026-06-01
ci: add always-on e2e-test-unit job (pytest tests/unit tests/utils)

aoshen02:
## What

Wires the **pytest-based** `tests/unit` + `tests/utils` suites into CI as a single dedicated, **always-on** job `e2e-test-unit`. Today these only run locally — nothing in any workflow executes them.

## Why a separate job (vs. folding into an existing one)

- Every existing job runs **one script per matrix entry** (`python tests/<file>` / `gpu_lock_exec.py`), never `pytest` on a directory.
- `tests/unit/backends/megatron_utils/update_weight/*` import `megatron.core` **at module load**, so they cannot run on the GitHub-hosted CPU `e2e-test-plugin-contracts` runner (no megatron there). They need the CI image.

So `e2e-test-unit` runs `pytest tests/unit tests/utils` **inside `inferactinc/public:vime-vllm-cu129-latest`** (ships megatron + torch) on the **self-hosted** runner, **0 GPUs**, **always-on** (no `run-ci-*` label, mirroring `e2e-test-plugin-contracts`). It does `needs: pre-commit` like the others.

## Coverage / risk

- `tests/unit` is **vime-original** (PRs #21/#22/#25) — covers the vLLM rollout/engine/arguments + colocate IPC weight-sync surface, i.e. the code most exposed to the ongoing slime→vime upstream sync.
- `tests/utils` is **mostly inherited from upstream slime** (#1204/#1742/#1776/#1866/#1874); only `test_vllm_config.py` is vime-adapted.
- Source-checked: **no** `cuda`/`.cuda()`/`device="cuda"`, no real `vllm`/`megatron` imports at test-file scope (megatron is transitive via the module-under-test). Pure CPU/mock → no `--gpus`, no GPU lock, no GPU contention with the other self-hosted jobs.

## Verification status (please confirm on the runner)

- ✅ Template edited + regenerated (`generate_github_workflows.py`); `pr-test.yml` is generated, not hand-edited.
- ✅ Generated YAML parses; 11 jobs; `check yaml` pre-commit hook passes.
- ✅ On a bare CPU box (no megatron): **149 passed**; the only failures were the 19 `update_weight` tests failing with `ModuleNotFoundError: megatron` — exactly what running **inside the image** fixes.
- ⏳ **Not yet executed inside the cu129 image** (the box I authored this on is arm64/cu13 and can't run that image). Please trigger via `workflow_dispatch` or a PR synchronize and confirm `e2e-test-unit` goes green; if the image lacks `pytest`, the job installs it.

## How to run

Always-on, so it fires on every PR `synchronize`/`labeled` and on manual dispatch:
```bash
gh workflow run pr-test.yml --repo vllm-project/vime --ref ci-unit-tests
```

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -616,6 +616,44 @@ jobs:
 
 
 
+  e2e-test-unit:
+    name: Unit & utils tests (pytest, in-image)
+    needs: pre-commit
+    # Always-on regression gate (like e2e-test-plugin-contracts): runs on every
+    # PR synchronize/label and on manual dispatch, no run-ci-* label required.
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
+    runs-on: self-hosted
+    defaults:
+      run:
+        working-directory: ${{ github.workspace }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Execute
+        shell: bash
+        run: |
+          # Run inside the CI image: it ships megatron + torch, which the
+          # tests/unit/backends/megatron_utils/update_weight tests need at
+          # import time. CPU/mock only -> no --gpus / no GPU lock.
+          docker run --rm \
+            --network host \
+            --ipc=host \
+            --shm-size=16g \
+            -e http_proxy \
+            -e https_proxy \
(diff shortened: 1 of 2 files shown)


CalvinXKY:
LGTM

PR #120, opened 2026-06-02
Add VLM Geo3k multiturn experiment

andakai:
## Summary

This PR fixes the Geo3K VLM multi-turn rollout path so that Geo3k multiturn with vLLM  reward score can increase.

### Geo3K VLM multi-turn rollout

- Reworked `examples/geo3k_vlm_multi_turn/rollout.py` to use the full-conversation vLLM render route as the source of truth for prompt tokens and multimodal features.
- Assistant tokens are appended from vLLM generation output.
- Tool/observation tokens are sliced from the next full render using a pending offset and masked out from loss.
- Added strict prefix-stability metadata/check support for detecting render-token drift.
- Appends a masked EOS token after stop-string generation when needed, and uses that appended EOS in the next observation offset.
- Propagates vLLM routed expert metadata through the custom rollout path.
- Supports local image paths and normalized `data:image/None` URLs for render requests.

### Geo3K environment behavior

- `examples/geo3k_vlm_multi_turn/env_geo3k.py` now ends an episode as soon as `calc_score` returns a correct answer.
- Wrong answers continue receiving feedback until `max_turns`.
- If the model does not produce a valid tool call, the env now extracts a final boxed/text answer, scores it, and ends the episode. This preserves reward for correct final answers that are not emitted in tool-call format.

### Run script defaults

- `examples/geo3k_vlm_multi_turn/run_geo3k_vlm_multi_turn.py` now uses the validated Qwen3-VL-2B colocated vLLM/Megatron setup:
  - `Qwen3-VL-2B-Instruct`
  - 4 GPUs
  - rollout TP = 1
  - Megatron TP = 1
  - `--vllm-router-policy round_robin`
  - `--vllm-max-model-len 32768`
  - `--vllm-gpu-memory-utilization 0.9`
  - `--vllm-generation-config vllm`
  - `--vllm-enforce-eager`
  - `--vllm-logprobs-mode processed_logprobs`
  - GRPO, lr `1e-6`, global batch size `512`

## Validation

W&B run:

- https://wandb.ai/danab2-hong-kong-university-of-science-and-technology/vime-dev/runs/pak7pj5q

image

## Why vLLM eager mode

Without `--vllm-enforce-eager`, Qwen3-VL multimodal rollout showed a large silent parity drift between vLLM rollout logprobs and Megatron replay logprobs. The VLM two-turn diagnostic reached about `0.1399` train/rollout logprob absolute diff.

With eager mode, the same path dropped to text-only-level parity, around `0.020` in smoke diagnostics and around `0.014-0.016` in the long training run. This matches SkyRL's stability-oriented default of using eager vLLM inference for its training engine.

cc @CalvinXKY


andakai:
- The enforce eager problem is fixed by https://github.com/vllm-project/vllm/pull/43617. vllm 0.22.0 doesn't have this pr included. I have tested locally: if using this pr, the logprob will become low. Will update this in the later vllm version.
- The router policy can be "consistent hash" instead of "round-robin". This will have higher prefix cache hit rate.

## Prefix Cache Hit Rate Comparison

I compared the vLLM engine-side `Prefix cache hit rate` logs for two 5-step no-enforce-eager Geo3K VLM multi-turn runs:

| Router policy | Log samples | Mean hit rate | Min | Max |
| --- | ---: | ---: | ---: | ---: |
| `consistent_hash` | 103 | **82.6%** | 60.9% | 90.7% |
| `round_robin` | 113 | **63.4%** | 46.5% | 77.4% |

`consistent_hash` improves the average engine-reported prefix cache hit rate by **19.2 percentage points** over `round_robin`, which is about a **30% relative improvement**.  The rollout time is similar due to the long-tail prompt.

image


Both runs completed successfully:

- `consistent_hash + no enforce eager`: https://wandb.ai/danab2-hong-kong-university-of-science-and-technology/vime-dev/runs/96i0vy4r
- `round_robin + no enforce eager`: https://wandb.ai/danab2-hong-kong-university-of-science-and-technology/vime-dev/runs/93mq3y9f

examples/geo3k_vlm_multi_turn/run_geo3k_vlm_multi_turn.py
```diff
@@ -3,7 +3,7 @@
 import vime.utils.misc as U
 from vime.utils.external_utils.command_utils import execute_train
 
-MODEL_NAME = os.environ.get("SLIME_SCRIPT_MODEL_NAME", "Qwen3-VL-2B-Instruct")
```
CalvinXKY: Remove "SLIME_SCRIPT_MODEL_NAME " from Readme. Consider refresh it too.

aoshen02: Will fix in this PR: https://github.com/vllm-project/vime/pull/121/changes#diff-c538ffca60ebb211024d65a44fbca17242c074e6e02504542ec09fa20cade232. pr 105 only rename the path.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -210,21 +285,33 @@ async def generate(args: Any, sample: Sample, sampling_params) -> Sample:
...
+                "sample_len": len(sample.tokens),
+                "rendered_len": len(rendered_ids),
+            }
+            if getattr(args, "strict_multiturn_render_token_match", False) and not is_prefix_stable:
+                raise RuntimeError(
+                    "Full conversation render is not prefix-stable with the generated token stream: "
+                    f"{sample.metadata['multiturn_render']}"
+                )
+
+        sample.multimodal_train_inputs = _multimodal_train_inputs_from_features(latest_features)
```
CalvinXKY: `sample.tokens` is stitched across turns, but multimodal features come from a single latest render. To assert that the count of `image_token_id` in `sample.tokens` matches `image_grid_thw.shape[0]` (or equivalent) before assigning `multimodal_train_inputs`, and fail loudly on mismatch rather than silently returning `None`. Is that better?

andakai: I tested and generally these two are the same. But there may be some mismatches in other scenarios, so I have added a validation here.

CalvinXKY (approved):
LGTM

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -188,13 +268,32 @@ async def generate(args: Any, sample: Sample, sampling_params) -> Sample:
...
+            if append_stop_eos:
+                stop_strings = (stop,) if isinstance(stop, str) else tuple(stop)
+                already_has_eos = bool(train_tokens and train_tokens[-1] == eos_token_id)
+                if stop_strings and response_text.endswith(stop_strings) and not already_has_eos:
+                    if getattr(args, "use_rollout_routing_replay", False):
+                        raise RuntimeError(
+                            "Routing replay is not supported when appending an artificial EOS after a stop string, "
+                            "because vLLM does not return routed experts for that extra token."
+                        )
+                    train_tokens.append(int(eos_token_id))
```
aoshen02, line 287: **#1 (correctness, multi-turn off-by-one) — part 1/2.** This synthetic EOS is appended to `train_tokens` only; it is *not* added to `messages` (L298 stores just `response_text`). So the next render's assistant segment has 1 fewer token than `len(train_tokens)`. See the offset note on L318.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -210,21 +309,35 @@ async def generate(args: Any, sample: Sample, sampling_params) -> Sample:
...
-
             if turn_idx + 1 >= args.max_turns:
-                sample.status = Sample.Status.COMPLETED
+                sample.status = Sample.Status.TRUNCATED
                 break
 
-        sample.multimodal_train_inputs = _merge_multimodal_train_inputs(multimodal_train_inputs_buffer)
+            next_user_message = env.format_observation(observation)
+            messages.append(next_user_message)
+            pending_obs_offset = len(input_ids) + len(train_tokens)
```
aoshen02: **#1 (correctness) — part 2/2.** `len(train_tokens)` includes the synthetic EOS appended at L290, but the re-rendered conversation (messages has only `response_text`, no EOS) has 1 fewer token before the observation. So next turn `obs_tokens = input_ids[pending_obs_offset:]` (L244) drops the first real observation token, desyncing `sample.tokens` from the context the model actually generated against.

Fix: compute the render-prefix length from `len(new_tokens)` (exclude the synthetic EOS), keep `train_tokens` separate:
```python
render_prefix_len = len(input_ids) + len(new_tokens)
pending_obs_offset = render_prefix_len
```
Only triggers when `append_stop_eos` fires.

aoshen02 (commented):
Code-review notes (Codex strict pass + manual). Inline comments below; 2 are correctness (multi-turn token alignment), the rest are simplifications. Pre-existing items not in this PR's diff (dead `_extract_balanced_json`, optional `orjson`, bare `env.close()` except, commented-out `eval_args`) are noted separately, not inlined since they predate this PR.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -210,21 +309,35 @@ async def generate(args: Any, sample: Sample, sampling_params) -> Sample:
...
+            rendered_body = await render()
+            latest_features = rendered_body.get("features")
```
aoshen02: **Cleanup (redundant).** This `latest_features = rendered_body.get("features")` is dead: the next loop iteration re-assigns it at the top (L241) from the same `rendered_body`, and the only consumer (L335) runs after a `break` that already saw L241's value. Safe to delete.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -210,21 +309,35 @@ async def generate(args: Any, sample: Sample, sampling_params) -> Sample:
...
+            rendered_ids = _coerce_flat_int_token_ids(rendered_body.get("token_ids"))
+            is_prefix_stable = rendered_ids[:pending_obs_offset] == sample.tokens[:pending_obs_offset]
+            sample.metadata["multiturn_render"] = {
+                "prefix_stable": is_prefix_stable,
+                "prefix_len": pending_obs_offset,
+                "sample_len": len(sample.tokens),
+                "rendered_len": len(rendered_ids),
+            }
+            if getattr(args, "strict_multiturn_render_token_match", False) and not is_prefix_stable:
```
aoshen02: **#2 (correctness) — the alignment guard is off by default.** The prefix-stability check at L322 is exactly what catches the #1 desync (and any render-vs-accumulation mismatch), but it only raises when `strict_multiturn_render_token_match` is True — which defaults to False. By default a mismatch is silently recorded in metadata and training proceeds on misaligned tokens.

Suggest defaulting strict to True (raise on mismatch), or making `rendered_ids` canonical and rebuilding `sample.tokens`/`loss_mask`/`rollout_log_probs` from the rendered conversation.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -146,38 +169,95 @@ async def generate(args: Any, sample: Sample, sampling_params) -> Sample:
...
+    if getattr(args, "router_policy", None) == "consistent_hash":
+        sample.session_id = sample.session_id or str(uuid.uuid4())
+        headers = {"x-session-id": sample.session_id}
+
+    build_env = env_module.build_env
+    if not callable(build_env):
+        raise ValueError("Environment module must expose a callable `build_env(sample, args)`.")
+    try:
+        env = build_env(sample=sample, args=args)
+    except TypeError:
```
aoshen02: **Cleanup.** This `except TypeError` retry can swallow a real `TypeError` raised *inside* `build_env` (a constructor bug), making debugging hard. `build_env` here accepts keywords — just call `env = build_env(sample=sample, args=args)`.

examples/geo3k_vlm_multi_turn/env_geo3k.py
```diff
@@ -224,6 +234,7 @@ def step(self, response_text: str):
 
         score = self._score_answer(parsed_answer)
         self.last_tool_score = score
+        self.correct = score == 1.0
```
aoshen02: **Cleanup (redundant state).** `self.correct` is only read immediately below at L249; it doesn't need to live on the env. A local `correct = score == 1.0` plus `return obs, correct or is_final_turn, info` removes the field and its init/reset at L42/L49.

examples/geo3k_vlm_multi_turn/run_geo3k_vlm_multi_turn.py
```diff
@@ -3,7 +3,7 @@
...
+MODEL_NAME = "Qwen3-VL-2B-Instruct"
 assert MODEL_NAME in {
```
aoshen02, line 7: **Cleanup (dead assert).** `MODEL_NAME` is a hardcoded constant immediately asserted to be in a set — the assert is always true. Either restore an env override (`MODEL_NAME = os.environ.get(...)`) or drop the assert and keep just the constant.

examples/geo3k_vlm_multi_turn/rollout.py
andakai: These two parts are fixed by separating render-vs-training lengths:

- Keep appending synthetic EOS to train_tokens as before.
- Compute the render prefix with original generation tokens only: render_prefix_len = len(input_ids) + len(new_tokens) and use that for pending_obs_offset.

examples/geo3k_vlm_multi_turn/rollout.py
andakai: True, this is redundant. Deleted

examples/geo3k_vlm_multi_turn/rollout.py
andakai: Yes this is false by default. Now I change it to fail fast whenever mismatch happens, without relying on "strict_multiturn_render_token_match" now.

examples/geo3k_vlm_multi_turn/rollout.py
andakai: Fixed.

examples/geo3k_vlm_multi_turn/env_geo3k.py
andakai: Fix. This is truly redundant.

examples/geo3k_vlm_multi_turn/run_geo3k_vlm_multi_turn.py
andakai: Fixed. Use the "VIME_SCIRPT_MODEL_NAME" now.

PR #126, opened 2026-06-02
chore: trim examples to qwen3 only

aoshen02:
This PR removes the public example set down to Qwen3-4B and Qwen3-30B-A3B, plus the related OPD and low-precision example/docs entries we already agreed to drop.

It is intentionally separate from the slime->vime rename PR so the rename can be rebased cleanly after this lands.


CalvinXKY (approved):
LGYM. Ensure it doesn't affect the CI.

PR #137, opened 2026-06-03
[F] simplify flush_cache: drop retry loop (vLLM /reset_prefix_cache always returns 200)

aoshen02:
## Summary

Originally this PR ported slime #1953 (log non-200 on `flush_cache` + back off). On review that change is **inert on vLLM**: vime's `flush_cache` posts to vLLM's `/reset_prefix_cache`, whose handler **always returns `Response(status_code=200)`** (it ignores the internal success/False bool) and gives no busy signal. So the 60× retry loop, the non-200 log, and the `TimeoutError` can never fire.

This PR instead **drops the loop entirely** and collapses `flush_cache` to a single best-effort POST + `raise_for_status()`.

## Why this diverges from slime #1953 (intentional, maintainer-approved)

slime's loop is a real drain barrier: sglang `/flush_cache` returns **400 while the engine has pending requests**, so slime retries until the engine drains. vLLM has no equivalent — `/reset_prefix_cache` 200s unconditionally, and vime's pause uses `mode="keep"` (it never uses flush as a drain barrier; colocate drains via `/sleep`). **SkyRL** (same vLLM keep-mode flow) likewise does a single `reset_prefix_cache` call with no loop.

## Diff

`flush_cache`: 22-line retry loop → single `POST /reset_prefix_cache?reset_running_requests=false&reset_external=false` + `raise_for_status()`. `py_compile` + `ruff` clean; `time` import still used elsewhere.

### DIFF
--- a/vime/backends/vllm_utils/vllm_engine.py
+++ b/vime/backends/vllm_utils/vllm_engine.py
@@ -786,22 +786,18 @@ class VLLMEngine(RayActor):
         return response
 
     def flush_cache(self):
-        """Clear prefix cache via ``POST /reset_prefix_cache``."""
+        """Reset the prefix cache via ``POST /reset_prefix_cache``.
+
+        vLLM's endpoint always returns 200 (it does not signal a busy engine),
+        so a single call suffices — no retry loop. ``reset_running_requests``
+        stays False: if any block is still in use the server skips the reset and
+        still returns 200; vime only calls this when the engine is idle
+        (colocate sleep / keep-mode pause).
+        """
         if self.node_rank != 0:
             return
-        params = {"reset_running_requests": False, "reset_external": False}
-        for _ in range(60):
-            try:
-                response = requests.post(f"{self._http_base()}/reset_prefix_cache", params=params, timeout=60)
-                if response.status_code == 200:
-                    return
-            except requests.ConnectionError:
-                raise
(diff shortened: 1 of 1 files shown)


vime/backends/vllm_utils/vllm_engine.py
```diff
@@ -773,6 +773,8 @@ def flush_cache(self):
                 response = requests.post(f"{self._http_base()}/reset_prefix_cache", params=params, timeout=60)
                 if response.status_code == 200:
                     return
+                logger.info("Error resetting vLLM prefix cache: HTTP %s %r", response.status_code, response.text)
```
CalvinXKY: logger.info("Error  -> logger.warning  is better?

aoshen02:
Dropping per maintainer call. #137's only change was logging a non-200 from `/reset_prefix_cache` + a 1s backoff — but vLLM's `/reset_prefix_cache` **always returns 200** (the handler ignores the internal success/False bool), so that branch is dead code on vLLM. SkyRL (same vLLM keep-mode practice) doesn't carry it either, and slime #1953's actual value (retry-until-drained, which relies on sglang's 400-while-busy) does not translate to vLLM's endpoint. flush_cache's existing loop + exception-retry on `main` is unaffected.

princepride (approved):
LGTM

PR #147, opened 2026-06-04
[Sync][G] streaming rollout + retool example (slime #1921/#1861)

aoshen02:
### Included
- **#1921 streaming** → `vime/rollout/vllm_streaming_rollout.py` + `tests/test_qwen3_4B_streaming_partial_rollout.py`. Translates the sglang streaming path to vLLM's `/inference/v1/generate` **SSE**. Load-bearing translation: sglang streams **cumulative** `output_token_logprobs` (overwrite per chunk); vLLM SSE sends **delta** `token_ids`+`logprobs` per chunk → accumulate (`+=`) and re-decode accumulated tokens each chunk (multi-byte-safe), finalize via `_vllm_meta_from_generate_choice`. `routed_experts` applied only from the terminal choice. MM raises `NotImplementedError` (the two-call render→generate flow doesn't fit a single stream). Streaming file is the reviewed `sync/slime-pr-1921` (draft #117) design.
- **#1861 retool** → re-added `examples/retool/generate_with_retool.py` (vime lacked `examples/retool/`) from the slime base + #1861's multi-turn/retry-desync hardening (`if not meta_info.get("output_token_logprobs")` guard after routing through the vLLM field map).

### Notes / follow-ups
- Retool example is **partial**: it depends on a sibling `tool_sandbox.py` not yet ported (out of #1861's scope; vime never had `examples/retool/`). The example won't run end-to-end until that lands — flagged for a follow-up.
- Two small correctness guards vs slime in retool (reviewed): post-loop status won't clobber a budget-driven `TRUNCATED`; `finish_type` tracked locally since vLLM responses are per-call.

### CI
- `test_qwen3_4B_streaming_partial_rollout` registered in the e2e-test-megatron matrix (`.j2` + generated `.yml`). py_compile clean.

---
## Fidelity ledger (audited 2026-06-04)

| op | detail |
|---|---|
| **Ported** | #1861 retool multi-turn/retry hardening ; #1921 streaming example |
| **Translated** (sglang→vllm) | streaming `meta_info` from sglang cumulative form → vLLM delta-accumulation |
| **Dropped/Deferred** | none |
| **Verified** | green on rebuilt FCBAGDE stack |


CalvinXKY (approved):
LGTM

PR #158, opened 2026-06-05
test: port test_value_temperature from slime #1928 (mirror-slime gap)

aoshen02:
## Summary
slime #1928 landed **both** the critic-temperature fix (`loss.py`) **and** `tests/test_value_temperature.py`. vime #95 ported only the fix — the test was missed. This restores it (mirror-slime).

- `tests/test_value_temperature.py`: slime verbatim, only `slime.`→`vime.` (import + `sys.modules` paths). `NUM_GPUS=0`, 0 sglang refs. **Passes locally (1 passed).**
- Wired into the `num_gpus:0` unit matrix in `pr-test.yml` + `pr-test.yml.j2`. vime CI runs explicit per-file matrix entries (not bare `pytest`), so an unlisted test would never run.

AI-assisted. Signoff item 3.


CalvinXKY (approved):
LGTM

PR #170, opened 2026-06-07
[Refactor] Align UpdateWeightFromTensor structure with slime (colocate weight sync)

aoshen02:
## What

Structural refactor of the colocate IPC weight-sync trainer (`UpdateWeightFromTensor`) to follow slime's layout line-for-line, with the vLLM-specific pieces isolated. Builds on #169. **−60 net lines.**

## Changes

**`connect_rollout_engines`**
- rename `_colocated_engines` → `rollout_engines`, `_distributed_engines` → `distributed_rollout_engines`; add the `use_distribute` flag
- collapse the `_ipc_engine_coordinator` / `_ipc_engine_slot_start` / `_ipc_engine_slot_end` / `_ipc_slot_group` quartet into `_ipc_gather_group` + `_ipc_gather_src` (slot leader == `rank == _ipc_gather_src`, `slot_size == get_world_size(group)`); guarded single-pass group creation
- drop the unused `rollout_engine_lock` store

**`update_weights`**
- pause / flush / continue act on `self.rollout_engines` only (the distributed pause/flush/continue is dropped — that path is dead under `args.colocate`, which is the only mode that selects this class; the distributed *send* stays, folded into `_send_hf_params`)
- add `torch.cuda.ipc_collect()` per chunk + after the loop (was missing — a likely producer-side IPC leak over many sync rounds)
- remove the per-chunk global barrier; restore the single after-loop barrier

**send path**
- `_send_hf_params` now returns `(refs, long_lived_tensors)`; `update_weights` does `ray.get(refs)` then `del long_lived_tensors`
- `_send_to_colocated_engine` returns `(refs, weight_refs)` with **no internal `ray.get` / per-slot barrier**. Synchronization is the `all_gather_object` collective deferral + the after-loop barrier, plus the receiver-side `torch.accelerator.synchronize()` already in the worker extension.

## Irreducible vLLM divergences (kept, encapsulated in `_send_to_colocated_engine`)
1. `reduce_tensor` UUID-keyed handles + `all_gather_object` + merge (vs slime's `FlattenedTensorBucket` + `gather_object`) — forced by the vLLM IPC format and the Megatron-reload monkey-patch (`gather_object` unsupported there).
2. vLLM #39212 `init` / `start` / `finish_weight_update` state machine (no sglang analog).

`pop_metrics` intentionally not added (unused anywhere in vime).

image

image
image

### DIFF
--- a/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
+++ b/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
@@ -15,11 +15,13 @@ from __future__ import annotations
 import os
 from argparse import Namespace
 from collections.abc import Callable, Iterable, Mapping, Sequence
+from typing import Any
 
 import ray
 import torch
 import torch.distributed as dist
 from megatron.core import mpu
+from ray import ObjectRef
 from ray.actor import ActorHandle
 
 from vime.utils.distributed_utils import get_gloo_group
@@ -123,24 +125,11 @@ def _merge_ipc_update_infos(infos: Sequence[dict[str, list]]) -> dict[str, list]
 
 
 class UpdateWeightFromTensor:
-    """Update colocated vLLM engines via CUDA IPC, with NCCL fallback for
-    non-colocated overflow engines. See the module docstring for the
-    high-level design (why we dispatch via ``update_weights_from_tensor``
-    directly instead of vLLM's ``trainer_send_weights``).
-
-    Engine lifecycle per ``update_weights`` call::
-
-        colocated:   pause_generation / flush_cache      (rank 0)
-        distributed: pause_generation / flush_cache      (rank 0)
-        init_weight_transfer_engine                      (rank 0, colocated, first call only)
-        start_weight_update                              (coordinator rank per engine only)
-        [for each HF chunk]
-          update_weights_from_tensor                     (per-rank or coordinator-merged)
-          update_weights_from_distributed                (src rank, distributed)
-          barrier                                        (all ranks)
-        finish_weight_update                             (coordinator rank per engine only)
-        colocated:   continue_generation                 (rank 0)
-        distributed: continue_generation                 (rank 0)
+    """
+    Update rollout engines from tensor dict:
+    gather TP(GPU NCCL) → convert HF(GPU) → send.
+    Colocated: build CUDA IPC handles → all_gather_object(Gloo CPU, over the engine
+    slot ranks) → Ray IPC to engine.  Distributed: GPU NCCL broadcast to remote engines.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -198,233 +178,209 @@ def connect_rollout_engines(
...
-            if self._colocated_engines:
-                ray.get([engine.pause_generation.remote() for engine in self._colocated_engines])
-                ray.get([engine.flush_cache.remote() for engine in self._colocated_engines])
-            if self._distributed_engines:
-                ray.get([engine.pause_generation.remote() for engine in self._distributed_engines])
-                ray.get([engine.flush_cache.remote() for engine in self._distributed_engines])
-            if self.quantization_config and self.quantization_config.get("quant_method") in ["compressed-tensors"]:
+            ray.get([engine.pause_generation.remote() for engine in self.rollout_engines])
+            ray.get([engine.flush_cache.remote() for engine in self.rollout_engines])
+            if self.quantization_config and self.quantization_config["quant_method"] in ["compressed-tensors"]:
```
knlnguyen1802, line 276: `self.quantization_config["quant_method"]` can cause key error, the old one `self.quantization_config.get("quant_method")` is safer

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -198,233 +178,209 @@ def connect_rollout_engines(
...
+        # vLLM #39212: exit weight-update mode.
+        if self._ipc_engine is not None and rank == self._ipc_gather_src:
             ray.get(self._ipc_engine.finish_weight_update.remote())
         dist.barrier(group=get_gloo_group())
 
-        # ── 6. Post-process quantization (if needed) and resume ───────────────
+        # int4/fp4 post_process
         if rank == 0:
-            if self.quantization_config and self.quantization_config.get("quant_method") in ["compressed-tensors"]:
+            if self.quantization_config and self.quantization_config["quant_method"] in ["compressed-tensors"]:
```
knlnguyen1802, line 312: Same problem

knlnguyen1802 (approved):
LGTM, we can refactor it to same as slime first, then can analyze the performance drop compare to slime. Give comment about minor issue

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
aoshen02: so as long as the model's huggingface config has quantization_config, it definitely has quant_method I think. I've run it and it works for me, so I think we can keep the same code with slime.

PR #182, opened 2026-06-08
examples: fix README drift and remove incomplete retool example

aoshen02:
## What

Continuing the docs/examples consistency pass. `examples/README.md` had drifted from the actual `examples/` contents in both directions, and one example directory was an incomplete stub.

## Changes

**`examples/README.md`**
- ❌ Removed dead link **`eval_multi_task`** — no such directory under `examples/`.
- ❌ Removed dead link **`reproducibility`** — it's a `docs/` page, not an example directory.
- ✅ Added **`coding_agent_rl`** — a complete example (README + `generate.py` + shared adapter) that existed on disk but was unlisted.

**Removed `examples/retool/`**
- It was an incomplete stub: only `generate_with_retool.py` + `tool_sandbox.py`, with **no README and no run scripts** (unlike upstream slime's full `retool/`). It was unlisted in the README and not referenced anywhere else in the repo (verified via grep).

## Verification

- After the change, every directory linked in `examples/README.md` exists, and every example directory on disk is listed.
- `grep -rin retool` across `*.py/*.sh/*.md/*.yml/*.j2` (excluding the deleted dir) returns no external references.

## Out of scope (inherited from slime — flagged, not touched here)

These are identical in upstream slime and fixing them in vime alone would diverge from the mirror:
- `examples/DrGRPO/custom_reducer.py` is cited in `customization.md` and `vime/utils/arguments.py` (as `Dr.GRPO`, with a spelling mismatch) but exists in neither vime nor slime.
- `examples/geo3k_vlm/README.md` has a broken self-link `./examples/geo_3k_vlm/run_geo3k_vlm_sft.sh` (typo), same as slime.


aoshen02:
Folded into #180 to keep all content-trim docs/examples changes in a single PR.

PR #232, opened 2026-06-10
sync: complete slime #1920, #1967, #1985

aoshen02:
## Summary

Traced 3 individual lint/cosmetic diffs back to their source slime PRs  and synced all remaining changes from each:

**slime #1985** (make tests shorter):
- Wrap `NamedTemporaryFile` across 3 lines in `test_vllm_config_mixed_offload_ft.py`
- Remove extra blank line in `test_vllm_config_mixed_offload.py`
- Parameter shortening already synced in vime PR #218

**slime #1920** (move fully_async example to main codebase):
- Rewrite `README.md` to match upstream (qwen2.5-0.5B, not qwen3-4b)
- Add `run-qwen2.5-0.5B-fully_async.sh` with proper vLLM translations
- Delete `run-qwen3-4b-fully_async.sh` (upstream removed it)

**slime #1967** (fix PYTHONBUFFERED typo):
- Fix `PYTHONBUFFERED=16` → `PYTHONUNBUFFERED=1` in 3 scripts: `run-glm4.7-30B-A3B.sh`, `run-glm4.7-355B-A32B.sh`, `run-minimax-m2.sh`
- `command_utils.py` already fixed in prior sync

## Test plan

- [ ] Verify no remaining `PYTHONBUFFERED` typos: `grep -r PYTHONBUFFERED .`
- [ ] Verify `run-qwen3-4b-fully_async.sh` is deleted
- [ ] Verify `run-qwen2.5-0.5B-fully_async.sh` has correct vLLM translations (pkill pattern, flag names)
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #233, opened 2026-06-10
refactor(arguments): align with slime + revert spurious dist.gather_object patch

aoshen02:
## Summary

- **`vime/backends/vllm_utils/arguments.py`** (415→361 lines): trim verbose docstrings/comments, slim `SKIPPED_DESTS` to names-only, add mutual-exclusion assertions in `validate_args` mirroring slime (`vllm_config` / `prefill_num_servers` / `rollout_external` conflict checks). `_make_add_argument_wrapper` stays public (5 unit tests reference it by name); `validate_args` intentionally omits `vllm_tp_size` (global TP broke heterogeneous per-group engines — per-engine resolver owns it, pinned by test).

- **`vime/utils/reloadable_process_group.py`**: drop `dist.gather_object` monkey-patch (added by PR #22, never in slime; unused in vime — callers explicitly use `all_gather_object` to stay on the patched path).

## Test plan

- [ ] `pytest tests/unit/backends/vllm_utils/test_arguments.py -m unit` — all argument contract tests pass
- [ ] `pytest tests/ -m unit` — full unit suite

### DIFF
--- a/vime/backends/vllm_utils/arguments.py
+++ b/vime/backends/vllm_utils/arguments.py
@@ -1,13 +1,8 @@
 """vLLM rollout backend argument definitions.
 
-- Wholesale-imports ``AsyncEngineArgs.add_cli_args(parser)`` with a wrapper that
-  prefixes every flag with ``--vllm-`` and every dest with ``vllm_``.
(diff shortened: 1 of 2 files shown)


aoshen02:
Merging into PR #232 instead.

PR #235, opened 2026-06-11
[Doc] Fix RTD language toggle paths

aoshen02:
Fix the Sphinx language toggle so it handles Read the Docs project translation URLs.

This updates the toggle logic to work with paths like /projects/vime/en/latest/ and /projects/vime/zh/latest/ instead of only the old GitHub Pages-style paths.


CalvinXKY:
Title says doc fix, but 4/5 files are not docs. Can we split?

PR #238, opened 2026-06-11
refactor(test): flatten CPU unit tests and centralize import stubs

CalvinXKY:
## Summary

- Remove `tests/unit/` tree; move tests to `tests/` and `tests/utils/` to match slime layout.
- Add `tests/_unit_stubs.py` for shared optional-dep stubs (replacing nested `conftest` import-time mocks).
- Register moved tests in `cpu-unittest` CI; set `pythonpath = [tests]` in pytest config.
- Write generated workflow YAML as UTF-8 (fixes `check-yaml` on Windows).

## Test plan

- [x] `pytest` on moved test files (134 passed locally)
- [x] `pre-commit run --all-files`
- [x] `run-ci-cpu-unittest` label on PR


.github/workflows/generate_github_workflows.py
```diff
@@ -21,7 +21,7 @@ def main():
         content = template.render()
 
         yaml_path = template_path.with_suffix("")
-        with open(yaml_path, "w") as f:
+        with open(yaml_path, "w", encoding="utf-8") as f:
```
aoshen02, line 24: What is this for?

vime/backends/vllm_utils/vllm_engine.py
```diff
@@ -56,7 +56,9 @@ def _response_json(response: requests.Response) -> dict:
     try:
         response.raise_for_status()
```
aoshen02, line 57: Do we need to change this?

.github/workflows/generate_github_workflows.py
CalvinXKY: > What is this for?

Without this change, running on Windows will error.

->  ：Write generated workflow YAML as UTF-8 (fixes check-yaml on Windows).

pyproject.toml
```diff
@@ -42,6 +42,7 @@ ignore = [
 # -vv will also display tests with duration = 0.00s
 addopts = "--verbose --pyargs --durations=0 --strict-markers"  # always add these arguments to pytest
 testpaths = ["./tests"]  # must be an explicit path to avoid importing another "tests" module
+pythonpath = ["tests"]  # bare imports like ``import _unit_stubs`` / ``import _cp_dist_helpers``
```
aoshen02: I think we can remove this as we have added sys.path in every test file.

vime/backends/vllm_utils/vllm_engine.py
CalvinXKY: Yes.
cpu-unittest runs on Python 3.10. Without the guard, _response_json crashes when attaching the response body to HTTPError. 
The test only asserts __notes__ when available; the production guard keeps 3.10 hosts from hitting the same bug on real HTTP errors.

pyproject.toml
CalvinXKY: Either drop pythonpath and add the same bootstrap to test_vllm_rollout.py, or keep pythonpath as the single mechanism and remove the duplicated inserts in utils.

.github/workflows/generate_github_workflows.py
aoshen02: But why we need to care about the need in windows as a rl framework.

CalvinXKY: > But why we need to care about the need in windows as a rl framework.

this isn’t about supporting Windows for RL training.

The change only affects generate_github_workflows.py when regenerating pr-test.yml locally. Without encoding="utf-8", the script can write non‑UTF‑8 bytes (e.g. em dash in comments) under a non‑UTF‑8 system default encoding. That breaks check-yaml once the file is committed, regardless of where CI runs.

So this is to make generated workflow YAML always UTF‑8 and pre-commit/CI-safe. Some of us do edit Python and run pre-commit on Windows, but the fix is about portable generator output, not Windows runtime support for vime.

vime/backends/vllm_utils/vllm_engine.py
aoshen02: We can just upgrade it to 3.11 I think. But why slime and miles doesn't have a test on this.

CalvinXKY: Slime’s sglang_engine.py also calls e.add_note() with no guard; it just doesn’t have a Py3.10 unit test on this path, so the same bug would show up there too.

Options: (1) bump cpu-unittest to 3.11 and revert the guard, (2) keep the guard, or (3) drop/weaken the test. I’d pick (1) or (2)  

which one?

CalvinXKY: 1

aoshen02:
buildkit access right now has some problem, but the ci test is pass, so we will just merge it.

PR #242, opened 2026-06-12
[Feature]  TransferQueue Integration for Rollout-to-Training

miracle0517:
## ✨ Summary

Introduce an **optional VIME TransferQueue data path** for transferring rollout data to training.  
When enabled via `--enable-vime-transfer-queue`, rollout batches are written directly into the TransferQueue. Megatron actor/critic workers then fetch their DP-local training data straight from TQ, bypassing the previous Ray ObjectRef rollout payload path.

---

## 🔧 What’s Changed

- **TransferQueueBridge** – A new central component that handles:
  - TQ initialization and client connections
  - Field conversion
  - Partition cleanup
  - Unified data fetch/write APIs

- **Rollout Workers** now publish normalized rollout batches directly into the TQ.

- **Megatron Actor Workers** now fetch rollout data from the TQ instead of relying on Ray ObjectRefs.

- **Local Training Schedule Preservation** – After fetching from TQ, the following fields are kept intact:
  - `global_batch_sizes`
  - `num_microbatches`
  - `micro_batch_indices`

- **Backpressure & Cleanup** – TQ staleness backpressure is enforced, and explicit partition cleanup runs once consumers have finished.

- **Extended Field Support** – Both tensor and non‑tensor rollout fields are supported, including multimodal inputs, metadata, prompts, routing replay data, and extra configured fields.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,6 +19,7 @@ ray[default]
 ring_flash_attn
 safetensors
 tensorboard
+TransferQueue>=0.1.8
 transformers
 vllm-router>=0.1.14
 wandb
--- a/scripts/run-qwen3-4B-grpo-tq.sh
+++ b/scripts/run-qwen3-4B-grpo-tq.sh
@@ -1,0 +1,162 @@
+#!/bin/bash
+
+# for rerun the task
+ray stop --force
+pkill -9 ray
+pkill -9 python
+sleep 3
+pkill -9 ray
+pkill -9 python
+
+set -ex
+
+# will prevent ray from buffering stdout/stderr
+export PYTHONUNBUFFERED=1
+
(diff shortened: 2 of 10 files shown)


aoshen02:
Hi, thanks for the contribution, is there any experiment results?

miracle0517:
> Hi, thanks for the contribution, is there any experiment results?

The test results are here: https://github.com/vllm-project/vime/issues/244

aoshen02:
This scale is not big enough, could you test with bigger vl models?

miracle0517:
> This scale is not big enough, could you test with bigger vl models?

We will conduct follow-up tests on multi-machine scenarios with the vl model to obtain performance data. Thank you.

aoshen02:
> > This scale is not big enough, could you test with bigger vl models?
> 
> We will conduct follow-up tests on multi-machine scenarios with the vl model to obtain performance data. Thank you.

Cool, thank you.

PR #252, opened 2026-06-15
fix(arguments): include DP in TP auto-compute default

aoshen02:
## Summary
- The `vllm_parse_args` default for `vllm_tensor_parallel_size` was computed as `rollout_num_gpus_per_engine // pp_size`, missing the `dp_size` divisor
- When `--vllm-data-parallel-size > 1` (required for expert parallelism via `--vllm-enable-expert-parallel`), the default TP was too large
- Align the formula with `_resolve_vllm_parallel_sizes` in `vllm_engine.py` which correctly uses `gpus // (pp * dp)`

## Example
With `--rollout-num-gpus-per-engine 8 --vllm-data-parallel-size 4 --vllm-pipeline-parallel-size 1`:
- **Before**: default TP = 8 // 1 = 8 (wrong)
- **After**: default TP = 8 // (1 * 4) = 2 (correct, matches engine launch)

The actual engine launch was unaffected because `_resolve_vllm_parallel_sizes` computed correctly at runtime. But the argparser default was inconsistent, which could confuse validation or logging code reading `args.vllm_tensor_parallel_size` before engine launch.

## Test plan
- [ ] Verify `--rollout-num-gpus-per-engine 8 --vllm-data-parallel-size 4` produces TP=2 in args
- [ ] Verify DP=1 (default) behavior unchanged: `--rollout-num-gpus-per-engine 4` still produces TP=4

### DIFF
--- a/tests/utils/test_vllm_arguments.py
+++ b/tests/utils/test_vllm_arguments.py
@@ -340,5 +340,52 @@ def test_parse_args_default_attribute_set_even_without_register(args_mod, monkey
     assert ns.vllm_tensor_parallel_size == 8
 
 
+@pytest.mark.unit
+def test_parse_args_tp_default_with_dp(args_mod, monkeypatch):
+    """TP auto-compute must divide by DP: TP = gpus / (PP * DP)."""
+    monkeypatch.setattr(args_mod, "add_vllm_arguments", lambda p: p)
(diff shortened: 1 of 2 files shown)


CalvinXKY (approved):
LGTM.  
Note: Buildkite #101 fail

PR #253, opened 2026-06-15
docker: upgrade base to vLLM 0.23.0, remove CUDA 13 build path

aoshen02:
## Summary
- Upgrade base image from `v0.22.0-cu129-ubuntu2404` to `v0.23.0-cu129-ubuntu2404`
- Remove all `ENABLE_CUDA_13` conditional build paths (cu13 apt headers, TE source build, triton source build, TMS_CUDA_MAJOR)
- Simplify TE to wheel-only install, cublas-dev to unconditional
- Remove `build-cu13` justfile target
- Adapt `vllm.patch` line numbers for 0.23.0

## Why remove CUDA 13?
cu129 nvcc already supports sm100/sm120 (Blackwell). The cu13 build path was unnecessary and caused failures: cu13 apt packages hijack `/etc/alternatives/cuda` from `cuda-12.9` to `cuda-13.0` (which has no nvcc), breaking TE CMake with `No CMAKE_CUDA_COMPILER could be found`.

## Test plan
- [x] Built successfully on gb200 (arm64), h200 (x86), gb300 (arm64)
- [x] All three confirmed `vllm.__version__ == '0.23.0'`
- [x] `vllm.patch` apply verified via `patch --dry-run` on 0.23.0 base
- [x] GLM-4.7-30B smoke passed with the new image on gb300

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,13 +1,10 @@
-ARG BASE_IMAGE=vllm/vllm-openai:v0.22.0-cu129-ubuntu2404
+ARG BASE_IMAGE=vllm/vllm-openai:v0.23.0-cu129-ubuntu2404
 FROM ${BASE_IMAGE}
 
 # ======================================== Arguments =============================================
 
 ARG PATCH_VERSION=latest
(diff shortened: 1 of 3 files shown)


CalvinXKY (approved):
LGTM

PR #257, opened 2026-06-16
fix(data): reuse stored multimodal_inputs in length filter

Meihan-chen:
## Summary

Prompt-length filtering crashes for any multimodal (VLM) dataset when `--apply-chat-template` is set. This makes the feature effectively unusable for VLM training/eval. This PR fixes the crash with a minimal change and adds regression tests.

## Repro

Run a VLM RL job (e.g. the geo3k VLM example) with `--apply-chat-template` and any flag that enables prompt-length filtering — e.g. `--rollout-max-context-len`, `--rollout-max-prompt-len`, or `--eval-max-prompt-len`. The run dies during `RolloutManager.__init__` (dataset construction):

```
File ".../qwen_vl_utils/vision_process.py", line 489, in extract_vision_info
    if isinstance(message["content"], list):
TypeError: string indices must be integers, not 'str'
```

## Root cause

In `filter_long_prompt` (`vime/utils/data.py`), the multimodal branch re-derived vision info from `sample.prompt`:

```python
multimodal_inputs = process_vision_info(sample.prompt, processor)
processor_output = processor(text=sample.prompt, **multimodal_inputs)
```

When `apply_chat_template=True`, `Sample.prompt` holds the **rendered string** (images collapsed into `<|image_pad|>` placeholders), not the conversation list.
`process_vision_info` (qwen_vl_utils) expects a list of message dicts; iterating the string yields single characters, and `char["content"]` raises the TypeError.

The correct multimodal inputs are already computed during dataset construction and stored in `Sample.multimodal_inputs`, so the recomputation is both wrong and redundant.

### Why it's easy to hit

`--rollout-max-context-len` is the natural way to cap vLLM's `max_model_len` (avoiding an over-large KV cache reservation). Setting it auto-derives `rollout_max_prompt_len` (arguments.py), which becomes the dataset's `max_length` and activates this filter — so capping context length on a VLM run trips the crash.

## Fix

Reuse the multimodal inputs already computed during dataset construction instead of recomputing them from the string prompt, routing them through the same `build_processor_kwargs` helper the rollout path uses so the token length measured here matches the real pipeline:

```python
processor_kwargs = build_processor_kwargs(sample.multimodal_inputs)
processor_output = processor(text=sample.prompt, **processor_kwargs)
```

## Tests

Added `tests/test_filter_long_prompt_multimodal.py` (CPU unit tests, `@pytest.mark.unit`):

- `test_filter_long_prompt_multimodal_reuses_stored_inputs` — reproduces the crash (asserts the buggy `process_vision_info` path is not taken) and verifies  correct length filtering for multimodal samples.
- `test_filter_long_prompt_mixed_text_and_multimodal` — a mixed text-only + multimodal dataset, confirming both branches filter independently and the text-only path is unaffected.

Both fail before the fix (TypeError) and pass after.


aoshen02:
Look good to me, could you also make a pr to slime?

PR #281, opened 2026-06-22
feat(npu): use vllm-ascend verified commit in Dockerfile.npu

Meihan-chen:
Port the Dockerfile.npu improvements onto the ascend branch while keeping the ascend-specific build (colocate patches, build-context npu_patch, and the vime clone from the ascend branch):
- drop the hardcoded VLLM_COMMIT arg and instead check out vllm to the vllm-ascend verified commit, then apply vllm.patch and install once
- install the vllm-router dependency

### DIFF
--- a/docker/Dockerfile.npu
+++ b/docker/Dockerfile.npu
@@ -6,7 +6,6 @@ FROM ${BASE_IMAGE}
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 WORKDIR /root
 
-ARG VLLM_COMMIT=9090368b650896bf5fc990c921df7eb4c20355a5
 ARG MEGATRON_COMMIT=3714d81d418c9f1bca4594fc35f9e8289f652862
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.npu
```diff
@@ -106,6 +107,11 @@ RUN pip install --no-build-isolation "nvidia-modelopt[torch]>=0.37.0" && \
     pip install --no-deps --no-build-isolation -e /root/Megatron-LM && \
     pip install --no-deps --no-build-isolation -e /root/MindSpeed
 
+# Install the Vime's dependency - vllm-router
+RUN pip install \
+      -i https://mirrors.aliyun.com/pypi/simple/ \
+      vllm-router>=0.1.14
```
CalvinXKY, line 113: This package is already in the requirements, so why install it separately here?

Meihan-chen: It's in requirements.txt, but the global pip index configured above doesn't provide vllm-router, so it needs an extra source

CalvinXKY (approved):
LGTM

PR #296, opened 2026-06-28
fix(rollout): abort vLLM rollout via delete-type /abort_requests

aoshen02:
## What

Replace the pause/resume-based rollout abort with a **delete-type** abort, so that under `--partial-rollout` the long tail is **truncated to partial** (and resumable next step) instead of either deadlocking or running to completion.

### DIFF
--- a/vime/rollout/vllm_rollout.py
+++ b/vime/rollout/vllm_rollout.py
@@ -15,6 +15,7 @@ import numpy as np
 import vllm_router  # noqa: F401 — ensures vllm-router is importable on startup
 from tqdm import tqdm
 
+from vime.backends.vllm_utils.server_control import abort_inflight_requests
 from vime.rollout.base_types import RolloutFnEvalOutput, RolloutFnTrainOutput
 from vime.rollout.filter_hub.base_types import MetricGatherer, call_dynamic_filter
 from vime.utils.async_utils import run
@@ -38,6 +39,9 @@ __all__ = ["generate_rollout", "get_model_url"]
 logger = logging.getLogger(__name__)
 
 _PROCESSOR_PROMPT_KEYS = {"input_ids", "attention_mask"}
+
+# Re-sweep interval while draining; bounds how long a late straggler can run.
+_ABORT_RESWEEP_INTERVAL_S = 3.0
 
 
 def _coerce_flat_int_token_ids(ids: Any) -> list[int]:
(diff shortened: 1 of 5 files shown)


vime/rollout/vllm_rollout.py
```diff
@@ -534,17 +538,15 @@ async def abort(args: Namespace, rollout_id: int) -> list[list[Sample]]:
...
-        pause_results = await asyncio.gather(*pause_tasks, return_exceptions=True)
-        for url, result in zip(urls, pause_results, strict=False):
-            if isinstance(result, Exception):
-                logger.warning(f"Failed to abort worker at {url}: {result}")
-        paused_workers = True
-
     count = 0
     while state.pendings:
-        done, state.pendings = await asyncio.wait(state.pendings, return_when=asyncio.FIRST_COMPLETED)
+        await abort_inflight_requests(urls)
```
Josephasafg, line 538: Do you think that in order to keep it less "wasteful" in terms of requests - maybe its worth calling `await abort_inflight_requests(urls)` once before the loop and then only when `asyncio.wait` times out?

vime/rollout/vllm_rollout.py
```diff
@@ -534,17 +538,15 @@ async def abort(args: Namespace, rollout_id: int) -> list[list[Sample]]:
...
-        logger.info(f"Abort request for {urls}")
-        pause_tasks = [post(f"{url.rstrip('/')}/pause?mode=abort", {}, max_retries=3) for url in urls]
-        pause_results = await asyncio.gather(*pause_tasks, return_exceptions=True)
-        for url, result in zip(urls, pause_results, strict=False):
-            if isinstance(result, Exception):
-                logger.warning(f"Failed to abort worker at {url}: {result}")
-        paused_workers = True
-
     count = 0
     while state.pendings:
```
Josephasafg: Optional safeguard: the resweep handles transient worker blips and multi-turn stragglers fine, but if an engine crashes/hangs (or the connection goes half-open) the aborted stream never returns and await post - `timeout=None` - hangs forever, so the drain re-sweeps indefinitely. Could we maybe do something like:


`_ABORT_MAX_DRAIN_S = 60`

in the main loop
```python
loop = asyncio.get_running_loop()
deadline = loop.time() + _ABORT_MAX_DRAIN_S
while state.pendings and loop.time() < deadline:
```

and then outside the loop:
```python
if state.pendings:
        for task in state.pendings:
            task.cancel()
        await asyncio.gather(*state.pendings, return_exceptions=True)
```

vime/rollout/vllm_rollout.py
aoshen02: fix

vime/rollout/vllm_rollout.py
aoshen02: I think for now we will keep parity with slime for this part. But will open to modify it if bug happens.

Josephasafg (approved):

Josephasafg:
LGTM. Thanks! I ran the use-case that failed for me and this fix seems to work.

CalvinXKY (approved):
LGTM

PR #317, opened 2026-07-03
fix(docker): /abort_requests abort-all fix + cu13 image variant

aoshen02:
## What

Fix the bundled vLLM `/abort_requests` endpoint in `docker/patch/latest/vllm.patch` so that **abort-all actually aborts**, and add a **cu13 Docker image variant** (partial port of #307).

## The bug

The endpoint (added in #296) handles an empty/missing `request_ids` by enumerating `engine.output_processor.request_states.keys()` and calling `engine.abort(request_ids)`.

But `request_states` is keyed by **internal** request ids, while `EngineClient.abort()` defaults to `internal=False` (treats ids as **external**, resolving them through the external→internal map). Under the default request-id randomization (`internal = f"{external}-{8 random}"`), the internal ids are not external keys, so the lookup matches nothing — `POST /abort_requests {}` returns **200 while aborting nothing**.

This was flagged by Codex on the upstream endpoint PR (vllm-project/vllm#47173, review r3516626381); the same code shipped here in the docker patch.

## The fix

Abort the all-in-flight list as **internal** (`internal=True`); keep external semantics for explicitly-passed `request_ids`. Also include `parent_requests` keys (parallel-sampling parents) and return **400** on malformed JSON. Mirrors vllm-project/vllm#47173.

## cu13 image variant

Partial port from #307 — excludes the NCCL_CUMEM default flip and glm5.2 scripts:

- `ENABLE_CUDA_13` build arg: cu13 dev headers, TE from source, skip cu12 cudnn pin
- `just build-cu13` + `cu13-*` tags via `just manifest cu13 ...`
- `TMS_CUDA_MAJOR` auto-detect from torch's CUDA version
- `git apply --allow-empty` so builds survive once the vllm patch is emptied upstream
- `actor_group`: preload `torch_memory_saver_hook_mode_preload_cu13.abi3.so`

## Testing

- Verified the patched `api_router.py` hunk applies cleanly to a pristine vLLM (`git apply --check`).
- The other patch sections (core.py, all2all_utils.py, disagg/protocol.py) are untouched.
- Full image build / runtime abort behaviour: via CI (Buildkite #430).

## AI assistance

Drafted with AI assistance (Claude); reviewed by a human before submission.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,19 +6,30 @@ FROM ${BASE_IMAGE}
 ARG PATCH_VERSION=latest
 ARG MEGATRON_COMMIT=1dcf0dafa884ad52ffb243625717a3471643e087
 
+ARG ENABLE_CUDA_13=0
+
 # ======================================== Setup =============================================
 
 WORKDIR /root/
 
 # ======================================== Apt dependencies =============================================
 
-# vllm/vllm-openai base is an inference image — add cu12 dev headers + cmake/git
+# vllm/vllm-openai base is an inference image — add CUDA dev headers + cmake/git
 # so TE / apex / flash-attn source builds find cusparse.h etc.
+# Dev packages must match the base toolkit: -12-9 for cu129, -13-0 for cu130.
 RUN apt-get update && apt-get install -y \
-    nvtop rsync dnsutils prometheus git cmake \
-    cuda-nvrtc-dev-12-9 cuda-nvml-dev-12-9 cuda-profiler-api-12-9 cuda-nvtx-12-9 \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -6,19 +6,30 @@ FROM ${BASE_IMAGE}
 ARG PATCH_VERSION=latest
 ARG MEGATRON_COMMIT=1dcf0dafa884ad52ffb243625717a3471643e087
 
+ARG ENABLE_CUDA_13=0
```
CalvinXKY, line 9: Scope note: this PR title only mentions abort fix, but commit 60b9ccf adds the full cu13 variant. Please expand the PR description (or split) so reviewers don't miss half the diff.

CalvinXKY (approved):

PR #329, opened 2026-07-08
[bugfix] Fix distributed update weights for pipeline parallel

knlnguyen1802:
## Summary

This PR fixes a hang in non-colocated full NCCL weight sync when the Megatron actor uses pipeline parallelism, for example `--tensor-model-parallel-size 1 --pipeline-model-parallel-size 2`.

The root cause is that Vime treated every `DP=0, TP=0` pipeline stage as a vLLM weight-sync source rank. With `PP>1`, multiple trainer ranks concurrently initialized separate trainer-to-vLLM NCCL transfer groups against the same vLLM rollout engines. Current vLLM weight transfer keeps one active receiver communicator, so later initialization could overwrite earlier initialization and leave trainer and receiver waiting on different NCCL groups.

The fix keeps the existing fast path for `PP=1`. For raw export with `PP>1`, Vime sends one pipeline stage at a time inside a single vLLM weight-update session. Megatron-Bridge has a different collective contract: every PP rank must participate in one export pass, while only the `DP=0, TP=0, PP=0` rank owns the vLLM receiver communicator and sends the complete HF weight stream.

## Problem

The default non-colocated full weight sync path is:

1. `MegatronTrainRayActor` selects `UpdateWeightFromDistributed` for `--update-weight-mode full --update-weight-transport nccl`.
2. `UpdateWeightFromDistributed.connect_rollout_engines` marks a trainer rank as a source when `data_parallel_rank == 0` and `tensor_model_parallel_rank == 0`.
3. Each source rank creates a trainer-to-vLLM NCCL transfer group through `connect_rollout_engines_from_distributed`.
4. Each bucket is sent through `NCCLWeightTransferEngine.trainer_send_weights`, while vLLM receives through its weight-transfer engine.

This works for `TP=2, PP=1` because there is only one `DP=0, TP=0` source rank.

It fails for `TP=1, PP=2` because both pipeline stages have `TP=0`:

| Global rank | TP rank | PP rank | Old source predicate |
| :--- | :---: | :---: | :---: |
| 0 | 0 | 0 | source |
| 1 | 0 | 1 | source |

The old code named those groups `vime-pp_0` and `vime-pp_1`, but the current vLLM-side wrapper does not actually use `group_name` to keep multiple receiver communicators. `VLLMEngine.init_weights_update_group` and `VLLMEngine.update_weights_from_distributed` pass the request to vLLM's weight-transfer endpoints, whose dense NCCL backend stores a single active `model_update_group` and receives from source rank `0`.

That makes the old multi-PP behavior unsafe:

1. PP stage 0 initializes receiver communicator A.
2. PP stage 1 initializes receiver communicator B.
3. vLLM remembers only the latest active receiver communicator.
4. A trainer rank can broadcast on communicator A while vLLM waits on communicator B.
5. NCCL waits forever because producer and consumer do not join the same group.

The rollout engine lock did not prevent this because it only serialized bucket broadcasts. It did not serialize or isolate communicator initialization.

## Goals

- Support non-colocated full NCCL weight sync when Megatron uses `PP>1`.
- Preserve existing behavior and performance for `PP=1`.
- Avoid requiring vLLM changes or a new receiver-side communicator map.
- Keep raw Megatron-to-HF and Megatron-Bridge export paths synchronized correctly.
- Make the implementation easy to reason about and test with CPU unit stubs.

### 1. Keep the PP=1 fast path unchanged

When `mpu.get_pipeline_model_parallel_world_size() == 1`, the implementation keeps the old persistent transfer group behavior:

1. The single `DP=0, TP=0` source rank connects rollout engines during `connect_rollout_engines`.
2. `update_weights` starts one vLLM weight-update session.
3. The source rank sends all buckets through the existing `_send_weights` path.
4. Vime synchronizes CUDA once after all buckets.
5. The vLLM weight-update session is finished.

This means the common non-PP case has the same process-group lifecycle and the same number of NCCL transfer groups as before.

### 2. Choose the communicator lifecycle by export mode

For raw export with `PP>1`, `connect_rollout_engines` records the rollout engines and defers group creation until each active pipeline stage is sent. This prevents multiple PP source ranks from racing to initialize the receiver communicator.

Megatron-Bridge export is collective across model-parallel ranks and yields the complete HF stream on every rank. It therefore creates one persistent communicator from `DP=0, TP=0, PP=0`; other PP ranks participate in export but never send.

### 3. Stage only the raw export path

`update_weights` still owns one outer vLLM lifecycle: pause, flush, start update, send weights, finish update, and resume.

For raw `PP>1`, the send loop activates one PP stage at a time. For Bridge `PP>1`, all ranks execute one synchronized Bridge export pass and PP0 sends the resulting buckets. This avoids both receiver-communicator races and Bridge collective deadlocks.

### 4. Preserve collective symmetry

The raw and Bridge paths have different contracts:

- Raw export is stage-local. Inactive PP stages skip local weight iteration but join the same raw-path barriers as the active stage.
- Megatron-Bridge export is collective. Every PP rank must call `get_hf_weight_chunks` in the same pass; only PP0 sends the complete exported stream to vLLM.

This distinction is required for correctness: making inactive Bridge ranks wait at a Vime barrier while PP0 enters Bridge collectives deadlocks before the first bucket is sent.

## Test Plan

Focused unit coverage should include:

- `PP>1` connect defers vLLM NCCL group initialization instead of creating a group immediately.
- The staged send loop initializes only the active PP stage's vLLM group.
- Inactive raw-path PP stages join two barriers without iterating local weights.
- Bridge export runs once on every PP rank, while only PP0 initializes the communicator and sends buckets.
- Existing packed and non-packed send behavior still calls `NCCLWeightTransferEngine.trainer_send_weights`.

Suggested commands:

```bash
pytest tests/utils/test_update_weight_from_distributed.py 
```

Runtime validation should include:

```bash
```

to confirm the existing non-PP path is unchanged, and:

```bash
--tensor-model-parallel-size 1
--pipeline-model-parallel-size 2
```

to confirm staged `vime-pp_0`, then `vime-pp_1`, sync completes without the NCCL hang.

When changing TP/PP topology between runs, use a topology-compatible checkpoint or start from model weights. Resuming a distributed optimizer checkpoint saved with a different `(TP, PP)` layout can fail before weight sync starts.

## Validation Results

Focused CPU tests:

```bash
PYTHONPATH=$PWD:/root/Megatron-LM \
  pytest -q tests/utils/test_update_weight_from_distributed.py \
            tests/test_qwen2_5_0_5B_non_colocate_pp.py
```

Result after the Bridge PP fix: `29 passed`.

A two-node integration run also exercised the staged path with:

- Trainer: 8 x H200, Megatron `TP=1, PP=8, EP=1`, BF16 master weights, torch distributed checkpoint.
- Rollout: 8 x H200, Qwen3-Coder-Next-FP8, two vLLM engines with `TP=4, EP=4` each.
- Result: all `vime-pp_0` through `vime-pp_7` stages synchronized, rollout resumed, and one training step completed.
- Correctness: `train/train_rollout_logprob_abs_diff = 0.011613311246037483`.

On this host pair, `mlx5_0` is RoCE while `mlx5_1..8` are InfiniBand. NCCL could otherwise select incompatible link types for different trainer GPUs. The integration run therefore set the same `NCCL_IB_HCA` selection on trainer and rollout nodes. This hardware routing requirement is independent of the PP serialization fix.

The E2E smoke test was also renamed to a valid Python module name (`test_qwen2_5_0_5B_non_colocate_pp.py`) so repository-wide pytest collection does not interpret the dots in the model name as package separators.

### Additional Bridge PP validation

The CI E2E was reproduced in an isolated container on the exact PR head plus this fix:

- Trainer: 2 x H200, Megatron `TP=1, PP=2`, Bridge export.
- Rollout: 2 x H200, two vLLM 0.24.0 engines with `TP=1` each.
- Initial sync: `3.1s`; subsequent syncs: `0.8s` and `0.9s`.
- Correctness: `train_rollout_logprob_abs_diff = 0.0068646` and `0.0083602`.
- Result: two rollout/train iterations completed and the Ray job succeeded.

The local H200 host required `NCCL_NVLS_ENABLE=0` because its Fabric Manager could not bind NVLS multicast memory; this is a host configuration workaround and is independent of the Bridge PP fix.


aoshen02:
I think we can add a customized (different from slime) gpu test set to do non colocate e2e test. Could you add one?

knlnguyen1802:
> I think we can add a customized (different from slime) gpu test set to do non colocate e2e test. Could you add one?

Ok let's me add one

knlnguyen1802:
cc @aoshen02

.buildkite/pipeline.yml
```diff
@@ -150,7 +150,7 @@ steps:
             value: short
           - label: "run-ci-vllm-config — 8 GPU, 4 tests"
             value: vllm-config
-          - label: "run-ci-megatron — 8 GPU, 14 runs"
```
aoshen02, line 153: Let's make a label called run-ci-vime-customized.

tests/test_qwen2_5_0_5B_non_colocate_pp.py
```diff
@@ -0,0 +1,112 @@
...
+import os
+
+import vime.utils.external_utils.command_utils as U
+
+
+MODEL_NAME = "Qwen2.5-0.5B-Instruct"
+MODEL_TYPE = "qwen2.5-0.5B"
+NUM_GPUS = 4
+
+
```
aoshen02, line 12: Why there are two tests?

vime/backends/megatron_utils/update_weight/update_weight_from_distributed.py
```diff
@@ -338,7 +393,8 @@ def _update_bucket_weights_from_distributed(
         ray.get(refs)
         converted_named_tensors.clear()
         ray.get(self.rollout_engine_lock.release.remote())
-        pbar.update(1)
+        if pbar is not None:
+            pbar.update(1)
```
aoshen02: Why modify this?

tests/test_qwen2_5_0_5B_non_colocate_pp.py
knlnguyen1802: It's light weight and enough to prove working of this fix

vime/backends/megatron_utils/update_weight/update_weight_from_distributed.py
knlnguyen1802: Let's me recover this

tests/test_qwen2_5_0_5B_non_colocate_pp.py
knlnguyen1802: Sorry there are duplicated test, let's me fix

aoshen02:
@codex review

PR #337, opened 2026-07-10
Forward recompute flags to the bridge provider; two hybrid model fixes

yuchenwang3:
Three small fixes needed to run a hybrid Mamba + attention + MoE model (NemotronH) through the megatron-bridge path.

- `model_provider.py`: the bridge-built provider only copies a fixed allowlist of args (TP/PP/CP/EP/...), so `--recompute-*` never reaches it and activation recompute is silently disabled. On a large hybrid MoE that leaves all expert activations resident and OOMs. Forward the recompute args so they take effect.

- `model.py`: `MambaModel.forward` has no `loss_mask` kwarg (GPTModel does), which raises at the first train step. Drop the kwarg when the model's forward does not accept it.

- `ray/rollout.py`: blank `PYTORCH_CUDA_ALLOC_CONF` for the vLLM engines. `expandable_segments:True` (set on the trainer to reduce fragmentation) breaks vLLM's custom all-reduce CUDA IPC, so engines fail to start unless the allocator stays at default.

### DIFF
--- a/vime/backends/megatron_utils/model.py
+++ b/vime/backends/megatron_utils/model.py
@@ -627,12 +627,17 @@ def train_one_step(
             # vime-patch: mcore MambaModel.forward (hybrid NemotronH) has no
             # loss_mask kwarg (GPTModel does). Drop it when unsupported; loss
             # masking happens in vime's own loss fn, not the model.
-            import inspect as _inspect
-
             _m = model
             while hasattr(_m, "module"):
                 _m = _m.module
-            if "loss_mask" not in _inspect.signature(_m.forward).parameters:
+            # Signature does not change during training, so compute once and cache.
+            accepts_loss_mask = getattr(_m, "_vime_forward_accepts_loss_mask", None)
+            if accepts_loss_mask is None:
+                import inspect
+
+                accepts_loss_mask = "loss_mask" in inspect.signature(_m.forward).parameters
+                _m._vime_forward_accepts_loss_mask = accepts_loss_mask
+            if not accepts_loss_mask:
                 forward_kwargs.pop("loss_mask", None)
 
             if batch["multimodal_train_inputs"] is not None:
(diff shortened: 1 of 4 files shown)


yuchenwang3:
Thanks for the review. Addressed all three: the forward-signature check is now computed once and cached on the model instead of per step, the recompute args use `getattr` with defaults, and the allocator config now only strips `expandable_segments` rather than clearing everything else.

aoshen02:
Could you take a look at this pr https://github.com/yuchenwang3/vime/pull/1? I think miles has already fix this problem, we can just migrate.

yuchenwang3:
Thanks @aoshen02 — merged yuchenwang3/vime#1 into this branch. The PR now carries miles' full `_apply_bridge_runtime_config()` (parallelism / numerics / recompute / offload / fp8 / MoE groups) instead of my field-by-field whitelist, so the two bridge integrations stay in sync. The `pg_collection` forwarding in `wrapped_bridge_provider` also fixes the PP>1 `provide()` path for hybrid Mamba models (NemotronH), which was the second fix here. Ready for another look.

aoshen02:
pre-commit failed, please go through the cpu test and we can enabled the gpu test here in CI

yuchenwang3:
@aoshen02 rebased onto today's main (the branch was old enough to carry a stale `.buildkite/pipeline.yml`, which is likely what tripped pre-commit). The diff is back to exactly the four intended files and `pre-commit run` passes on all of them locally (ruff / autoflake / isort / black). CPU test should be good to go — please re-trigger when you get a chance.

aoshen02:
> @aoshen02 rebased onto today's main (the branch was old enough to carry a stale `.buildkite/pipeline.yml`, which is likely what tripped pre-commit). The diff is back to exactly the four intended files and `pre-commit run` passes on all of them locally (ruff / autoflake / isort / black). CPU test should be good to go — please re-trigger when you get a chance.

I just did it lol

PR #178, opened 2026-06-08
refactor(rollout): align vllm_rollout.py with slime sglang_rollout.py

aoshen02:
## Summary

- Mirror slime sglang_rollout.py structure in vllm_rollout.py (984→756 lines, 29→16 top-level defs)
- Remove 13 vime-only helper functions not present in slime — inlined where needed, removed where over-defensive
- Add get_rollout_num_engines() in http_utils.py to mirror slime (replaces inline rollout_num_gpus // rollout_num_gpus_per_engine)
- Add no_stop_trim → include_stop_str_in_output mapping in _build_inference_sampling_params (subsumes PR #71)
- Remove no-op return_routed_experts from request payload (vLLM is engine-level --enable-return-routed-experts, not per-request like SGLang)
- Remove dead output_token_logprobs construction (update_from_meta_info never reads it)

### Removed functions (no slime equivalent)

| Function | Lines | Reason |
|---|---|---|
| _base_dataset_prompt_ids | 13 | vime-only partial-continuation budget; slime does not have it |
| _apply_vllm_routed_experts | 41→7 | Collapsed to inline block mirroring slime |
| _inference_generate_tokens_and_logprobs | 30→6 | Simplified inline |
| _align_engine_tokens_and_logprobs | 13 | Over-defensive padding |
| _warn_if_sampling_filters_may_affect_logprobs | 9 | Noise warning |
| _vllm_meta_from_generate_choice | 18 | Inlined into generate() |
| _decode_vllm_routed_experts | 3 | Inlined |
| _router_worker_urls | 9 | Inlined into abort() |
| _resume_vllm_workers | 10 | Inlined into abort() |

### Kept (vLLM-specific, no sglang equivalent)

- _coerce_flat_int_token_ids — vLLM /inference/v1/generate requires flat list[int]
- _build_inference_sampling_params — sglang to vllm param translation
- _mm_render_response_to_generate_body + _find_token_subsequence + _align_mm_feature_placeholders_to_tokens — vLLM MM uses render+generate two-step (vs sglang direct image_data)

### sglang to vllm API divergences (intentionally NOT mirrored)

- return_routed_experts: SGLang supports per-request, vLLM is engine-level — removed from payload
- routed_experts location: SGLang returns in output meta_info, vLLM returns in output choices[0]
- abort: SGLang uses POST /abort_request, vLLM uses POST /pause?mode=abort + POST /resume
- multimodal: SGLang sends image_data directly, vLLM needs render+generate two-step

[...]

PR #341, opened 2026-07-11
[Bugfix][Rollout] Fix Geo3K VLM multi-turn rollout

### LINKED ISSUE #331: [Bug] geo3k_vlm_multi_turn example failed with import error
### Bug Description

When I run this `geo3k_vlm_multi_turn` example, I encountered the following error:
```sh
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/home/vime-proj/vime/examples/geo3k_vlm_multi_turn/rollout.py", line 15, in <module>
    from vime.rollout.vllm_rollout import (
ImportError: cannot import name '_apply_vllm_routed_experts' from 'vime.rollout.vllm_rollout' (/home/vime-proj/vime/vime/rollout/vllm_rollout.py)
```
It seems to relate with this refactor pr #178 which deleted the required functions imported in this example.

### Steps to Reproduce

Just run this script: examples/geo3k_vlm_multi_turn/run_geo3k_vlm_multi_turn.py

### Expected Behavior

No bug occurs.

### Actual Behavior

`ImportError: cannot import name '_apply_vllm_routed_experts' from 'vime.rollout.vllm_rollout' (/home/vime-proj/vime/vime/rollout/vllm_rollout.py)`

Same with `_inference_generate_tokens_and_logprobs` func

### Environment

- vime version: main
- Python version: 3.11
- PyTorch version: 2.10.0

### Pre-submission Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/vllm-project/vime/blob/main/CONTRIBUTING.md) and understand the collaboration scope.
- [x] I have read the [documentation](https://vllm-project.github.io/vime/) and my issue is not addressed there.
- [x] I have searched for [existing issues](https://github.com/vllm-project/vime/issues) and this is not a duplicate.
- [x] I have provided a minimal, reproducible example.

---
I tried to add these functions back, and this time I met the following error:
```sh
File "/usr/local/python3.12.13/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/home/vime-proj/vime/vime/rollout/vllm_rollout.py", line 697, in generate_rollout_async
    group: list[Sample] = task.result()
                          ^^^^^^^^^^^^^
  File "/home/vime-proj/vime/vime/utils/trace_utils.py", line 374, in async_wrapper
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vime-proj/vime/vime/rollout/vllm_rollout.py", line 587, in generate_and_rm_group
    group = await asyncio.gather(*tasks)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vime-proj/vime/vime/utils/trace_utils.py", line 374, in async_wrapper
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vime-proj/vime/vime/rollout/vllm_rollout.py", line 523, in generate_and_rm
    sample = await custom_generate_func(args, sample, sampling_params)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vime-proj/vime/examples/geo3k_vlm_multi_turn/rollout.py", line 327, in generate
    raise RuntimeError(
RuntimeError: Full conversation render is not prefix-stable with the generated token stream: {'prefix_stable': False, 'prefix_len': 697, 'sample_len': 6
97, 'rendered_len': 753}
```

Can you re-verify this example?

FeathBow:
## Summary

Fixes #331.

- Remove the stale `_apply_vllm_routed_experts` and `_inference_generate_tokens_and_logprobs` imports left behind by #178.
- Render the initial multimodal prompt once and reuse the returned vLLM features for later turns.
- Preserve generated token IDs as the canonical conversation history and append only the chat-template observation suffix between turns.
- Add a deterministic single-GPU Geo3K custom-rollout E2E regression to the manual Buildkite `short` suite.

#178 removed or inlined two private helpers from `vime.rollout.vllm_rollout`, but the Geo3K example continued importing them, causing the module-level `ImportError` reported in #331.

Restoring those helpers alone was insufficient. The legacy implementation decoded the generated assistant response and re-rendered the complete conversation before every turn. Chat-template rendering can tokenize that decoded response differently from the exact token stream returned by vLLM, which caused the prefix-stability failure reported in #331.

The fixed path renders only the initial prompt. Later turns preserve the exact generated token IDs and append only the EOS-aware observation suffix.

## Validation

The E2E uses Qwen/Qwen3-VL-2B-Instruct, VeraIsHere/geo3k_imgurl_processed, one dataset row, two deterministic turns, and one GPU, against a real vLLM worker and router with one multimodal render request and two generation requests.

It verifies:

- public dotted-path loading of the Geo3K `generate` entry point
- exactly `render -> generate -> generate`
- first-turn generated token IDs are preserved verbatim in the second request
- both generation requests carry the render-derived multimodal features
- response tokens, loss masks, and rollout log probabilities remain aligned

Manual A/B validation reproduced the original `ImportError` and demonstrated the legacy re-render divergence: with the deleted helpers restored, the second-turn re-rendered request contained a duplicated EOS token that is absent from the vLLM-generated token stream, while the fixed path issues the canonical request. The rollout patch was applied to a fresh worktree at `b929921` (`origin/main`) and passed the single-GPU E2E; the vLLM worker, router, and engine processes were gone afterward. A later test-only cleanup removed redundant assertions and inlined test configuration without changing the rollout implementation.

## Scope

This test exercises the Geo3K model/data custom-rollout path. It is not a full training-step E2E and uses a deterministic two-turn test environment rather than the real `env_geo3k` behavior. Its assertions assume the pinned model's greedy trajectory completes two turns within the token budget on the CI image.

### DIFF
--- a/examples/geo3k_vlm_multi_turn/rollout.py
+++ b/examples/geo3k_vlm_multi_turn/rollout.py
@@ -1,23 +1,25 @@
 from __future__ import annotations
 
+import base64
 import importlib
 import importlib.util
+import io
 import json
 import sys
 import uuid
+from collections.abc import Mapping
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
+import numpy as np
 import torch
 from PIL import Image
 
-# When executed as a module: python -m examples.vlm_multi_turn.rollout
 from vime.rollout.vllm_rollout import (
     GenerateState,
-    _apply_vllm_routed_experts,
     _build_inference_sampling_params,
     _coerce_flat_int_token_ids,
-    _inference_generate_tokens_and_logprobs,
     _mm_render_response_to_generate_body,
 )
 from vime.utils.http_utils import post
@@ -25,102 +27,89 @@ from vime.utils.processing_utils import encode_image_for_rollout_engine
 from vime.utils.types import Sample
 
 DEFAULT_ENV_MODULE = "examples.geo3k_vlm_multi_turn.env_geo3k"
+DUMMY_MESSAGES = (
+    {"role": "system", "content": "You are a helpful assistant."},
+    {"role": "user", "content": "I am a user."},
+    {"role": "assistant", "content": "I am an assistant."},
+)
+IMAGE_GRID_DIMENSIONS = 3
 
 
 def _load_env_module(env_path: str | None):
-    """Load the interaction environment module from a module path or a file path."""
     target = env_path or DEFAULT_ENV_MODULE
     module_path = Path(target)
-    if module_path.suffix == ".py" and module_path.exists():
-        spec = importlib.util.spec_from_file_location(f"rollout_env_{module_path.stem}", module_path)
-        if spec is None or spec.loader is None:
-            raise ImportError(f"Cannot import environment module from {module_path}")
-        module = importlib.util.module_from_spec(spec)
-        sys.modules[spec.name] = module
-        spec.loader.exec_module(module)
-        return module
-    return importlib.import_module(target)
+    if module_path.suffix != ".py" or not module_path.exists():
+        return importlib.import_module(target)
+
+    spec = importlib.util.spec_from_file_location(f"rollout_env_{module_path.stem}", module_path)
+    if spec is None or spec.loader is None:
+        raise ImportError(f"Cannot import environment module from {module_path}")
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[spec.name] = module
+    spec.loader.exec_module(module)
+    return module
 
 
 def _image_to_render_url(image: Any) -> str:
-    """Convert a dataset/processor image object to the image_url payload accepted by vLLM render."""
-    if isinstance(image, str):
-        if image.startswith("data:"):
-            # Some prepared Geo3K rows use data:image/None; vLLM expects a concrete media type.
-            return image.replace("data:image/None;", "data:image/png;", 1)
-        if image.startswith(("http://", "https://")):
-            return image
-        image_path = Path(image).expanduser()
-        if image_path.exists():
-            with Image.open(image_path) as loaded_image:
-                return encode_image_for_rollout_engine(loaded_image)
+    if not isinstance(image, str):
+        return encode_image_for_rollout_engine(image)
+    if image.startswith("data:"):
+        return image.replace("data:image/None;", "data:image/png;", 1)
+    if image.startswith(("http://", "https://")):
+        return image
+
+    image_path = Path(image).expanduser()
+    if not image_path.exists():
         raise ValueError(f"Unsupported image string for vLLM render: {image!r}")
-    return encode_image_for_rollout_engine(image)
+    with Image.open(image_path) as loaded_image:
+        return encode_image_for_rollout_engine(loaded_image)
 
 
-def _build_initial_messages(sample: Sample) -> list[dict]:
-    """Build the initial conversation from the dataset prompt.
-
-    The preferred path mirrors SkyRL: keep the untemplated conversation as the source of truth
-    and let vLLM render/chat-template it on every turn. If an old pre-rendered string prompt is
-    supplied, fall back to pairing it with sample.multimodal_inputs while removing literal
-    <image> placeholders to avoid double image tokens.
-    """
+def _build_initial_messages(sample: Sample) -> list[dict[str, Any]]:
     if isinstance(sample.prompt, list):
         return [dict(message) for message in sample.prompt]
 
-    content: list[dict] = []
-    images = (sample.multimodal_inputs or {}).get("images") or []
-    for image in images:
+    content: list[dict[str, Any]] = []
+    for image in (sample.multimodal_inputs or {}).get("images") or []:
         content.append({"type": "image", "image": image})
-    # Backward-compatible fallback for old runs that pass a raw text prompt containing <image>.
-    text_prompt = str(sample.prompt).replace("<image>", "").lstrip()
-    content.append({"type": "text", "text": text_prompt})
+    content.append({"type": "text", "text": str(sample.prompt).replace("<image>", "").lstrip()})
     return [{"role": "user", "content": content}]
 
 
-def _messages_for_render(messages: list[dict]) -> list[dict]:
-    """Normalize per-turn messages to the render-route shape (image → image_url)."""
-    out: list[dict] = []
-    for msg in messages:
-        content = msg.get("content")
+def _messages_for_render(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    rendered: list[dict[str, Any]] = []
+    for message in messages:
+        content = message.get("content")
         if not isinstance(content, list):
-            out.append(msg)
+            rendered.append(dict(message))
             continue
-
-        rendered_content: list[dict] = []
+        parts: list[dict[str, Any]] = []
         for part in content:
             if part.get("type") == "image" and part.get("image") is not None:
-                rendered_content.append(
-                    {"type": "image_url", "image_url": {"url": _image_to_render_url(part["image"])}}
-                )
+                parts.append({"type": "image_url", "image_url": {"url": _image_to_render_url(part["image"])}})
             else:
-                rendered_content.append(part)
-        out.append({"role": msg["role"], "content": rendered_content})
-    return out
+                parts.append(dict(part))
+        rendered.append({"role": message["role"], "content": parts})
+    return rendered
 
 
-def _multimodal_train_inputs_from_features(features: Any) -> dict | None:
-    """Decode the latest vLLM render features into train-side multimodal tensors."""
+def _multimodal_train_inputs_from_features(features: Any) -> dict[str, torch.Tensor] | None:
     if not features:
         return None
-    if isinstance(features, str):
-        try:
-            features = json.loads(features)
-        except json.JSONDecodeError:
-            return None
-    if not isinstance(features, dict):
+    decoded = json.loads(features) if isinstance(features, str) else features
+    if not isinstance(decoded, dict):
+        raise TypeError(f"vLLM features must decode to a dictionary, got {type(decoded).__name__}")
+    kwargs_data = decoded.get("kwargs_data")
+    if not isinstance(kwargs_data, dict) or "image" not in kwargs_data:
         return None
-    kwargs_data = features.get("kwargs_data")
-    if not isinstance(kwargs_data, dict):
-        return None
-    if "image" not in kwargs_data:
-        return None
+    encoded_images = kwargs_data["image"]
+    if not isinstance(encoded_images, list):
+        raise TypeError("vLLM features.kwargs_data.image must be a list")
 
     from vllm.entrypoints.serve.disagg.mm_serde import decode_mm_kwargs_item as vllm_decode
 
     parts_by_key: dict[str, list[torch.Tensor]] = {}
-    for encoded in kwargs_data["image"]:
+    for encoded in encoded_images:
         item = vllm_decode(encoded)
         for key, value in item.get_data().items():
             if not isinstance(value, torch.Tensor):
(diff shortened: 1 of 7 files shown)


examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -1,153 +1,164 @@
...
 )
 from vime.utils.http_utils import post
 from vime.utils.processing_utils import encode_image_for_rollout_engine
 from vime.utils.types import Sample
 
 DEFAULT_ENV_MODULE = "examples.geo3k_vlm_multi_turn.env_geo3k"
+DUMMY_MESSAGES = (
+    {"role": "system", "content": "You are a helpful assistant."},
+    {"role": "user", "content": "I am a user."},
+)
```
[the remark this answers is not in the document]
FeathBow: Makes sense. Done!

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -158,187 +169,394 @@ def _validate_multimodal_train_inputs(sample: Sample, tokenizer: Any, processor:
...
-
-            current_sampling_params = sampling_params_for_turn()
-            if current_sampling_params is None:
-                sample.status = Sample.Status.TRUNCATED
-                break
+        eos_index = len(prefix_ids) - 1 - prefix_ids[::-1].index(eos_token_id)
+    except ValueError as exc:
+        raise ValueError(f"Dummy observation prefix lacks eos_token_id={eos_token_id}") from exc
+    boundary = prefix_ids[eos_index:] + rendered_ids[len(prefix_ids) :]
+    return boundary[1:] if canonical_ids and canonical_ids[-1] == eos_token_id else boundary
```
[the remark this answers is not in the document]
FeathBow: Thanks for flagging this. The suggestion rests on the premise that prefix_ids[-1] is the turn terminator — but that doesn't hold for the Qwen templates this example targets. Each Qwen turn renders as ...<|im_end|>\n, so the last prefix token is the newline separator (id 198), not <|im_end|> (id 151645). Verified with the pinned Qwen3-VL-2B tokenizer.

This matters because the boundary is also responsible for closing the assistant turn when the generated stream doesn't already end with EOS (e.g. a stop_token_ids finish). In that case the current code emits <|im_end|>\n<|im_start|>user…, while the suggested code would emit \n<|im_start|>user… — the assistant turn is never terminated, silently corrupting the training stream.

For templates whose turn terminator differs from eos_token_id, the current behavior is an explicit ValueError rather than silent mis-splicing, which is intentional fail-fast; this example is scoped to Qwen-family templates (the E2E pins Qwen3-VL).

FeathBow:
I've kept this PR scoped to the Geo3K regression reported in #331. While working on it I noticed the same pattern elsewhere, which may be worth a separate repository-wide follow-up: `examples/multi_agent/agent_system.py` still imports the removed `_inference_generate_tokens_and_logprobs`, and `examples/mem_agent/rollout_client.py` imports both that helper and the removed `_align_engine_tokens_and_logprobs`, so loading either example currently fails with the same kind of `ImportError`.

There are also several independent implementations of the vLLM `choices[0].token_ids` / `logprobs.content` parsing contract in the tree (at least four non-streaming plus a streaming-specific variant), and their malformed/missing-logprob behavior differs slightly, so consolidation seems better handled as a separate API-design discussion than as an incidental refactor in this bugfix. Happy to open a follow-up issue for either point if maintainers think it would be useful.

tests/test_geo3k_vlm_multi_turn_e2e.py
```diff
@@ -0,0 +1,281 @@
...
+import shlex
+import subprocess
+import sys
+import time
+from argparse import Namespace
+from collections.abc import Iterator
+from contextlib import contextmanager
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
```
aoshen02, line 16: Could you add a specific test suite vime-customized and add the test to it?

FeathBow: Thanks! makes sense. Done.

aoshen02:
could you resolve the conflict?

FeathBow:
Done, rebased with main

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -158,187 +170,394 @@ def _validate_multimodal_train_inputs(sample: Sample, tokenizer: Any, processor:
...
+        self.sample.response_length = 0
+        self.render_body = body
+        self.max_response_budget = _response_budget(
+            self.sampling_params,
+            self.args.rollout_max_context_len,
+            len(prompt_ids),
+        )
+
+    async def _generate_turn(self, sampling_params: dict[str, Any]) -> _Turn:
+        body = copy.deepcopy(self.render_body)
```
andakai: deepcopy may increase the memory use. I think `.copy` is enough.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -1,153 +1,165 @@
...
+
+
+def _merge_feature_tensors(parts_by_key: dict[str, list[torch.Tensor]]) -> dict[str, torch.Tensor]:
+    merged: dict[str, torch.Tensor] = {}
+    for key, values in parts_by_key.items():
+        merged[key] = values[0] if len(values) == 1 else torch.cat(values, dim=0)
+    return merged
+
+
+def _multimodal_train_inputs_from_features(features: Any) -> dict[str, torch.Tensor] | None:
```
andakai: Why replace the previous codes with these three seperate helper functions? I think the previous code can be more clear. I think there is no need to be defensive here and can be merged to `_multimodal_train_inputs_from_features()`

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -158,187 +170,394 @@ def _validate_multimodal_train_inputs(sample: Sample, tokenizer: Any, processor:
...
+            raise ValueError(
+                "vLLM routed experts shape does not match the generated sequence: "
+                f"actual_size={decoded.size}, expected_size={expected_size}"
+            )
+    return {"routed_experts": decoded}
+
+
+def _response_budget(sampling_params: dict[str, Any], context_limit: int | None, prompt_length: int) -> int | None:
+    if prompt_length < 0:
+        raise ValueError(f"prompt_length must be non-negative, got {prompt_length}")
```
andakai: no need to do this. the input is already `len()`

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -158,187 +170,394 @@ def _validate_multimodal_train_inputs(sample: Sample, tokenizer: Any, processor:
...
+        _validate_multimodal_train_inputs(
+            self.sample,
+            self.state.tokenizer,
+            self.state.processor,
+            mm_inputs=mm_inputs,
+        )
+        self.sample.multimodal_train_inputs = mm_inputs
+        self.sample.response = self.state.tokenizer.decode(self.response_tokens, skip_special_tokens=False)
+        if self.sample.status == Sample.Status.PENDING:
+            self.sample.status = Sample.Status.COMPLETED
```
andakai: When reaching the `finalize()`, its status will not be PENDING.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -158,187 +170,394 @@ def _validate_multimodal_train_inputs(sample: Sample, tokenizer: Any, processor:
...
+    reserved = {"tokenize", "add_generation_prompt"}.intersection(kwargs)
+    if reserved:
+        raise ValueError(f"template kwargs cannot override reserved keys: {sorted(reserved)}")
+    prefix = tokenizer.apply_chat_template(
+        list(DUMMY_MESSAGES),
+        tools=tools,
+        tokenize=True,
+        add_generation_prompt=False,
+        **kwargs,
+    )
```
andakai: no need to do `reserved` here. The override keys will be checked `tokenizer.apply_chat_template` here.

andakai (commented):
Thanks for your effort. The fix is correct. The refactor looks also generally good to me. The previous implementation basically refer to slime's design. Now the multi-turn's design is clear in the `_Geo3kRollout.run()`.  These two features also look great.
> Render the initial multimodal prompt once and reuse the returned vLLM features for later turns.
> Preserve generated token IDs as the canonical conversation history and append only the chat-template observation suffix between turns.

Here are some suggestions. If you have time, could you also run an e2e test and paste a reward curve to check? I can also do it later.

FeathBow:
Thanks for the detailed review! I've adopted all five inline suggestions and will push them shortly — the single-GPU E2E regression still passes after the changes. My cluster is tied up with other tasks at the moment, so the reward-curve run may take a while on my side.

FeathBow:
Btw it may be worth a separate follow-up for the remaining private-helper imports and duplicated token/logprob parsing. I’ve kept it out of scope for #341, but I’d be happy to track this if you think that would be useful :)

> I've kept this PR scoped to the Geo3K regression reported in #331. While working on it I noticed the same pattern elsewhere, which may be worth a separate repository-wide follow-up: `examples/multi_agent/agent_system.py` still imports the removed `_inference_generate_tokens_and_logprobs`, and `examples/mem_agent/rollout_client.py` imports both that helper and the removed `_align_engine_tokens_and_logprobs`, so loading either example currently fails with the same kind of `ImportError`.
> 
> There are also several independent implementations of the vLLM `choices[0].token_ids` / `logprobs.content` parsing contract in the tree (at least four non-streaming plus a streaming-specific variant), and their malformed/missing-logprob behavior differs slightly, so consolidation seems better handled as a separate API-design discussion than as an incidental refactor in this bugfix. Happy to open a follow-up issue for either point if maintainers think it would be useful.

FeathBow:
Update: I ran the example's training loop end-to-end with the fixed rollout on a single GPU (scaled config: 60 rollout steps, rollout batch 8 × 8 samples, --vllm-enforce-eager). Training is stable throughout: the truncation ratio drops from ~0.5–0.6 early on to ~0.2, and the smoothed raw reward ends the run at its high (~0.45–0.5). Curves below.
followtiraw reward
to outtruncale

aoshen02:
> Update: I ran the example's training loop end-to-end with the fixed rollout on a single GPU (scaled config: 60 rollout steps, rollout batch 8 × 8 samples, --vllm-enforce-eager). Training is stable throughout: the truncation ratio drops from ~0.5–0.6 early on to ~0.2, and the smoothed raw reward ends the run at its high (~0.45–0.5). Curves below. followtiraw reward to outtruncale

Hi, I wonder why use --vllm-enforce-eager, any blocker?

FeathBow:
Oh no new blocker from my side — I kept the example's stock flags for the run (only scaled num-rollout / batch sizes).

FeathBow:
--vllm-enforce-eager comes from the example itself `run_geo3k_vlm_multi_turn.py` note that "vLLM 0.22.0 needs eager mode for Qwen3-VL logprob parity", to be dropped once vllm#43617 lands. Since 43617 shipped in vLLM v0.23.0 , the flag could be removable now.

aoshen02:
> --vllm-enforce-eager comes from the example itself `run_geo3k_vlm_multi_turn.py` note that "vLLM 0.22.0 needs eager mode for Qwen3-VL logprob parity", to be dropped once vllm#43617 lands. Since 43617 shipped in vLLM v0.23.0 , the flag could be removable now.

Please remove it cc @andakai

FeathBow:
Done. If useful, I can also run a short training without the flag on my side.

andakai:
Yes, can remove `--vllm-enforce-eager` now.

By the way, I will also run an experiment tonight. Below is the previous curve. Compared with yours, I think we need to run a more obvious increase curve, which should basically align with the previous one. Because the current pr is a big refactor, I think we need to ensure its accuracy is correct.
image

FeathBow:
Sounds good — I'll also arrange a run tonight if my cluster permits.

andakai:
I have run a test. The reward curve seems good. Thanks for your effort. @FeathBow cc @aoshen02 

image

aoshen02:
The pre-commit failed.

FeathBow:
I'll check it.

FeathBow:
There was only one code formatting issue, fixed.

PR #332, opened 2026-07-10
refactor(vllm): use native IPC engine for colocated weight sync

miracle0517:
## Summary

Replace VIME's custom colocated IPC weight-sync implementation with the native weight transfer engines provided by vLLM and vLLM-Ascend.

This removes the manual IPC handle gathering, serialization, and worker-side chunk loading path while aligning colocated weight updates with vLLM's native `/update_weights` lifecycle.

## What changed

- Use  `NPUIPCWeightTransferEngine` on Ascend to send colocated weights.
- Route native `update_info` payloads through `VLLMEngine.update_weights` and vLLM's `/update_weights` endpoint.
- Initialize the native weight transfer engine once when colocated rollout engines are connected.
- Use the native `start_weight_update` / `finish_weight_update` state machine with checkpoint-format loading.
- Remove the custom:
  - trainer-rank IPC gather groups
  - IPC handle merging and serialization helpers
  - `update_weights_from_tensor` / `update_weights_chunk` RPC path
  - colocated worker-side tensor reconstruction logic
- Retain the Ascend worker compatibility patches for MoE expert weight loaders and rotary embeddings.
- Update unit tests to cover the native CUDA/NPU IPC transfer flow and isolate optional dependency stubs between test modules.

## Why

The previous implementation duplicated functionality already supported by vLLM's native IPC transfer engines and required VIME to maintain custom GPU-slot gathering, IPC handle reconstruction, and worker RPC code.

Using the native transfer path reduces VIME-specific integration code and keeps colocated weight synchronization aligned with upstream vLLM and vLLM-Ascend behavior.

## Testing

- [x] `pytest tests/unit/backends/megatron_utils/update_weight/test_update_weight_from_tensor.py -q`
- [x] `pytest tests/unit/backends/vllm_utils/test_vllm_engine.py -q`
- [x] Pre-commit formatting and lint checks
- [x] End-to-end Qwen3-4B colocated training on Ascend NPU

PR #345, opened 2026-07-13
refactor(vllm): use native IPC engine for colocated weight sync

miracle0517:
## Summary

1.Replace VIME's custom colocated IPC weight-sync implementation with the native weight transfer engines provided by vLLM and vLLM-Ascend.
This removes the manual IPC handle gathering, serialization, and worker-side chunk loading path while aligning colocated weight updates with vLLM's native /update_weights lifecycle.

2.Fix fused MoE weight shape mismatches during layerwise reload on Ascend NPU.

## What changed

- Use  `NPUIPCWeightTransferEngine` on Ascend to send colocated weights.
- Route native `update_info` payloads through `VLLMEngine.update_weights` and vLLM's `/update_weights` endpoint.
- Initialize the native weight transfer engine once when colocated rollout engines are connected.
- Use the native `start_weight_update` / `finish_weight_update` state machine with checkpoint-format loading.
- Remove the custom:
  - trainer-rank IPC gather groups
  - IPC handle merging and serialization helpers
  - `update_weights_from_tensor` / `update_weights_chunk` RPC path
  - colocated worker-side tensor reconstruction logic
- Retain the Ascend worker compatibility patches for MoE expert weight loaders and rotary embeddings.
- Update unit tests to cover the native CUDA/NPU IPC transfer flow and isolate optional dependency stubs between test modules.

## Why

1.The previous implementation duplicated functionality already supported by vLLM's native IPC transfer engines and required VIME to maintain custom GPU-slot gathering, IPC handle reconstruction, and worker RPC code.
Using the native transfer path reduces VIME-specific integration code and keeps colocated weight synchronization aligned with upstream vLLM and vLLM-Ascend behavior.

2.vLLM-Ascend may reload fused MoE weights with a transposed layout. Direct copying causes `npu_grouped_matmul` shape errors, while skip wake up transpose can solve it.

## Testing

- [x] `pytest tests/unit/backends/megatron_utils/update_weight/test_update_weight_from_tensor.py -q`
- [x] `pytest tests/unit/backends/vllm_utils/test_vllm_engine.py -q`
- [x] Pre-commit formatting and lint checks
- [x] End-to-end Qwen3-4B colocated training on Ascend NPU
- [x] `pytest tests/unit/backends/megatron_utils/update_weight/test_update_weight_from_tensor.py -q`
- [x] End-to-end Qwen3-30B-A3B colocated training on Ascend NPU

Comparison of reward and train_rollout_logprob_abs_diff curves between colocated and non-colocated training：
image
image

### DIFF
--- a/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
+++ b/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
@@ -1,27 +1,18 @@
 """
-Colocated vLLM weight sync (trainer + worker)
-=============================================
-
-Trainer: ``UpdateWeightFromTensor`` — Megatron → HF chunks → CUDA IPC (Ray).
-
-Worker: ``vLLMColocateWorkerExtension`` — passed to ``vllm serve`` via
-``--worker-extension-cls``; patches IPC receive before handle deserialisation.
-
-https://docs.vllm.ai/en/stable/examples/rl/rlhf_ipc/
+Colocated vLLM weight sync using native IPC transfer engines.
 """
 
 from __future__ import annotations
 
 import os
 from argparse import Namespace
-from collections.abc import Callable, Iterable, Mapping, Sequence
-from typing import Any
+from collections.abc import Callable, Mapping, Sequence
+from dataclasses import asdict
 
 import ray
 import torch
 import torch.distributed as dist
 from megatron.core import mpu
-from ray import ObjectRef
 from ray.actor import ActorHandle
 
 from vime.utils.common import is_npu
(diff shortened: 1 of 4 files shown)


CalvinXKY:
Title/body only cover the MoE transpose, but commit 38f431a is a large native-IPC refactor. Please expand the description so reviewers don’t miss the weight-sync rewrite. 
And close #332.

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -36,169 +27,6 @@
 )
 
 
-def _patch_npu_colocate_worker() -> None:
-    """Skip layerwise_reload in NPUWorker weight-update hooks (colocate OOM fix)."""
```
CalvinXKY, line 40: Please call out in the PR that this intentionally re-enables native layerwise on NPU (previously skipped for OOM). Any memory regress vs the old skip path?

CalvinXKY:
With native IPC, rank0 now fans out one merged-UUID payload to all colocated engines. Confirmed OK for multi-engine colocate, not only single-engine 30B?

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
miracle0517: > Please call out in the PR that this intentionally re-enables native layerwise on NPU (previously skipped for OOM). Any memory regress vs the old skip path?

We conducted a comparative experiment on Qwen3-30B. The peak memory usage of the new version compared to the old version is 35.51 GB vs. 34.89 GB, with a difference within 1.78%

miracle0517:
> With native IPC, rank0 now fans out one merged-UUID payload to all colocated engines. Confirmed OK for multi-engine colocate, not only single-engine 30B?

uuid is generated in npu_ipc_engine( physical_npu_id = npu_generate_uuid() = f"{get_ip()}-{physical_device}"),   so it is unique. And it is tested ok in multi-engine colocate。

CalvinXKY:
@miracle0517 @momo609

momo609:
LGTM

momo609 (approved):

Meihan-chen (commented):
TODO / follow-up (post-merge, non-blocking). This PR bundles a common-path refactor together with the NPU delta, which breaks the ascend fork's is_npu isolation and makes the eventual trunk merge harder to separate.

- The switch to vLLM's native /update_weights engine — VLLMEngine.update_weights, removal of the custom update_weights_chunk path, and the is_checkpoint_format default flip — is not NPU-specific; CUDA benefits too. It should land as its own trunk PR rather than ride along here.
- Only the genuinely NPU-specific parts (NPUIPCWeightTransferEngine selection, MoE transpose skip in wake_up, rotary patch) belong behind is_npu.
- Rule of thumb: if removing the is_npu guard would change CUDA behavior, it's a common change and belongs on trunk.

Will track as a follow-up item; flagging here so it isn't lost.

PR #346, opened 2026-07-14
[Megatron][Bridge] Port miles' full runtime-config forwarding to bridge provider

aoshen02:
## Summary

#337 fixed one dropped field (`recompute_*`) in the bridge path's args→provider copy in `model_provider.py`. Comparing against `miles`' equivalent (`miles/backends/megatron_utils/model_provider.py`), that whitelist is much narrower than what miles already forwards — this PR ports miles' `_apply_bridge_runtime_config()` helper wholesale so the two integrations stay in sync, instead of re-discovering each dropped field one bug report at a time.

- Replaces the inline whitelist with `_apply_bridge_runtime_config(provider, args)`, ported from miles, grouped by category (parallelism / loss-sequence / numerics / recompute / offload / comm-overlap / fp8 / attention backend / MoE) with miles' original comments.
- Also ports miles' `pg_collection` forwarding: PP>1 paths in some `megatron.bridge` providers (e.g. `mamba_provider`, needed for hybrid Mamba/attention models like NemotronH — the same model family #337 targets) read `self._pg_collection.pp` during `provide()`. Without forwarding it, those paths hit `AttributeError`. vime didn't have this at all.
- Unifies vime's separate actor/critic bridge-provider closures into one `wrapped_bridge_provider` so both get the same `config`/`pg_collection` handling (miles doesn't split actor/critic for the bridge path at all; this keeps vime's critic support while still getting the same fix).
- Kept vime's own `gradient_accumulation_fusion` field, which miles doesn't forward.

## Caveat — please verify before merging

This environment has no GPU / `megatron` / `megatron.bridge` importable, so I could only verify statically (syntax, `ruff`, `black`) plus cross-checking field names against vime's own existing unconditional usages elsewhere in the repo:
- `calculate_per_token_loss`, `deterministic_mode`, `tp_comm_overlap` are already accessed unconditionally elsewhere in `vime/backends/megatron_utils/` (`loss.py`, `initialize.py`), so these are confirmed safe.
- `attention_softmax_in_fp32`, `fp32_residual_connection`, `cpu_offloading_num_layers`, `distribute_saved_activations`, `fp8`, `fp8_recipe`, `attention_backend` are standard `megatron.training.arguments` fields per miles' usage, but I could not independently confirm they resolve on `args` in this environment — please run this through CI / a real bridge training config before merging.

## Testing

- `python -m ast` parse, `ruff check`, `black --check` on the changed file (all pass)
- No GPU/CI run performed here — relying on this repo's CI + reviewer's own hardware for the actual bridge path

### DIFF
--- a/vime/backends/megatron_utils/model.py
+++ b/vime/backends/megatron_utils/model.py
@@ -627,12 +627,17 @@ def train_one_step(
             # vime-patch: mcore MambaModel.forward (hybrid NemotronH) has no
             # loss_mask kwarg (GPTModel does). Drop it when unsupported; loss
             # masking happens in vime's own loss fn, not the model.
-            import inspect as _inspect
-
             _m = model
             while hasattr(_m, "module"):
                 _m = _m.module
-            if "loss_mask" not in _inspect.signature(_m.forward).parameters:
+            # Signature does not change during training, so compute once and cache.
+            accepts_loss_mask = getattr(_m, "_vime_forward_accepts_loss_mask", None)
+            if accepts_loss_mask is None:
+                import inspect
+
+                accepts_loss_mask = "loss_mask" in inspect.signature(_m.forward).parameters
+                _m._vime_forward_accepts_loss_mask = accepts_loss_mask
+            if not accepts_loss_mask:
                 forward_kwargs.pop("loss_mask", None)
 
             if batch["multimodal_train_inputs"] is not None:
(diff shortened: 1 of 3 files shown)


aoshen02:
Re-targeting: this should be a PR against #337's own branch (yuchenwang3:yuchen/forward-recompute-and-hybrid-fixes), not directly against main. Closing this one and opening the correctly-targeted PR instead.

PR #356, opened 2026-07-16
ci(rocm): AMD ROCm GPU CI on Buildkite

indianspeedster:
Adds an AMD ROCm CI path for vime on Buildkite.

- `.buildkite/pipeline-rocm.yaml` — declarative ROCm pipeline behind a manual block gate; each test runs the prebuilt ROCm image via docker and arbitrates GPUs with `gpu_lock_exec` on `HIP_VISIBLE_DEVICES`.
- Test-execution path: `U.is_rocm()` gate + a ROCm branch in `execute_train` (runs the train script directly against the ray head).
- `test_qwen2.5_0.5B_fully_async_short.py` adapted for the ROCm path (HF→Megatron convert, no bridge).

short suite grading (validated end-to-end on gfx950 / 8×MI350X):
- `fully_async` (0.5B) — gating, must pass; currently green.
- `gsm8k` / `gsm8k_async` (0.8B) — soft-fail, non-blocking: vLLM rollout returns NaN logprobs on ROCm, diverging the train-vs-rollout logprob check. Kept visible (orange) until fixed, then flipped to blocking.

Buildkite agent provisioning for queue `amd_gfx950` is complete; the suite runs on the self-hosted MI350X node.

### DIFF
--- a/tests/test_qwen2.5_0.5B_fully_async_short.py
+++ b/tests/test_qwen2.5_0.5B_fully_async_short.py
@@ -12,21 +12,18 @@ existing 0.5B short tests.
 import os
 import vime.utils.external_utils.command_utils as U
 
+
 MODEL_NAME = "Qwen2.5-0.5B-Instruct"
 MODEL_TYPE = "qwen2.5-0.5B"
 NUM_GPUS = 4
 
-HF_PATH = f"/root/models/{MODEL_NAME}"
-# ROCm converts HF->Megatron (no modelopt bridge). Write to a container-local
-# path so concurrent short tests sharing this model don't race on the output.
-MG_PATH = f"/tmp/{MODEL_NAME}_torch_dist"
-
 
 def prepare():
     U.exec_command("mkdir -p /root/models /root/datasets")
-    U.exec_command(f"hf download Qwen/{MODEL_NAME} --local-dir {HF_PATH}")
+    U.exec_command(f"hf download Qwen/{MODEL_NAME} --local-dir /root/models/{MODEL_NAME}")
     U.hf_download_dataset("zhuzilin/dapo-math-17k")
     if U.is_rocm():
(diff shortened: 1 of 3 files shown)


tests/test_qwen2.5_0.5B_fully_async_short.py
```diff
@@ -78,8 +93,8 @@ def execute():
 
     vllm_args = (
         "--rollout-num-gpus-per-engine 1 "
-        "--vllm-gpu-memory-utilization 0.65 "
-        "--vllm-max-cudagraph-capture-size 16 "
+        f"--vllm-gpu-memory-utilization {'0.3' if U.is_rocm() else '0.65'} "
```
aoshen02: why we need to change the utilization here

tests/test_qwen2.5_0.5B_fully_async_short.py
```diff
@@ -12,20 +12,35 @@
 import os
 import vime.utils.external_utils.command_utils as U
 
-
 MODEL_NAME = "Qwen2.5-0.5B-Instruct"
 MODEL_TYPE = "qwen2.5-0.5B"
 NUM_GPUS = 4
 
+HF_PATH = f"/root/models/{MODEL_NAME}"
```
aoshen02: Let's try to not modify this.

tests/test_qwen2.5_0.5B_fully_async_short.py
indianspeedster: Good catch. The lower utilization was a carry-over from colocated runs with larger models, I reverted it to 0.65 and validated on MI350x.

tests/test_qwen2.5_0.5B_fully_async_short.py
indianspeedster: Sure.

PR #357, opened 2026-07-16
[Bugfix][Rollout] Fix IPv6 vLLM engine and health-check URLs

ShuoleiWang:
## Summary

Fix IPv6 URL construction in the vLLM engine control plane.

## Problem

The vLLM subprocess requires an unbracketed IPv6 address when binding
the server, while an IPv6 literal must be enclosed in brackets when
used as the host component of an HTTP URL.

Previously, the health-check path could construct an invalid URL such
as:

`http://2001:db8::1:8000/health`

instead of:

`http://[2001:db8::1]:8000/health`

The health check may then keep retrying failed requests, making engine
startup appear to be stuck.

## Changes

- Reuse the existing `_wrap_ipv6` helper when constructing vLLM
  control-plane URLs.
- Keep the subprocess bind address unbracketed.
- Preserve the existing behavior for IPv4 and already-bracketed IPv6
  addresses.
- Add CPU unit coverage for engine initialization and health-check URL
  construction.

## Testing

- `pytest tests/utils/test_vllm_engine.py -q`
- `pre-commit run --all-files`

### DIFF
--- a/tests/utils/test_vllm_engine.py
+++ b/tests/utils/test_vllm_engine.py
@@ -478,6 +478,64 @@ def test_get_url_ipv6_host(vllm_engine):
 
 
 @pytest.mark.unit
+@pytest.mark.parametrize(
+    ("host", "expected_host"),
+    [
+        ("127.0.0.1", "127.0.0.1"),
+        ("2001:db8::1", "[2001:db8::1]"),
+        ("[2001:db8::1]", "[2001:db8::1]"),
+    ],
+)
+def test_init_formats_server_host_for_urls(vllm_engine, monkeypatch, host, expected_host):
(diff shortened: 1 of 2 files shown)


aoshen02:
interesting

ShuoleiWang:
> interesting

@aoshen02  Thanks for reviewing this PR! 
I have access to some H20 machines in my spare time. Could you recommend any specific areas where I could contribute to Vime?

aoshen02:
> > interesting
> 
> @aoshen02 Thanks for reviewing this PR! I have access to some H20 machines in my spare time. Could you recommend any specific areas where I could contribute to Vime?

You can ping me on vllm slack channel or add my wechat 18102550020?

PR #362, opened 2026-07-22
[Feature] Add DAPO preset and Soft Overlong support

CalvinXKY:
## Summary
- Expose `--advantage-estimator dapo` as a preset over the existing GRPO implementation.
- Enable Clip-Higher, token-level loss, the default dynamic-sampling filter, and Soft Overlong defaults.
- Add `--soft-overlong-cache` and apply its length penalty before existing reward normalization.
- Document the preset in English and Chinese.

## Design
DAPO does not add a second advantage, loss, or normalization path. Argument validation expands it to the existing GRPO primitives; the only new rollout behavior is Soft Overlong reward shaping.

## Test plan
- [x] Pre-commit on all changed files
- [x] `pytest tests/test_megatron_argument_validation.py` (18 passed)
- [x] Local DAPO preset smoke validation
- [x] Author GPU/NPU smoke evidence in PR comments


CalvinXKY:
NPU  tests:
image

CalvinXKY:
## GPU tests:

- Qwen3-4B 
- vime (commit 5cabf1f)
- vllm/vime:latest


image

image

vime/utils/arguments.py
```diff
@@ -1889,9 +1896,29 @@ def vime_validate_args(args):
...
 
+    if args.advantage_estimator == "dapo":
+        args.advantage_estimator = "grpo"
+        if args.eps_clip_high is None:
+            args.eps_clip_high = 0.28
+        args.calculate_per_token_loss = True
+        if args.dynamic_sampling_filter_path is None:
+            args.dynamic_sampling_filter_path = (
+                "vime.rollout.filter_hub.dynamic_sampling_filters.check_reward_nonzero_std"
+            )
```
aoshen02, line 1907: I think most of the ability of dapo is already support in vime, and for the soft overlong cache, I would say we recommend just use the --custom-reward-post-process-path instead of merging it, I'm still ok with merging this amount of code if it is needed.

CalvinXKY: OK.  Let's leave this as the solution for now.

PR #372, opened 2026-07-27
[NPU] fix: update MoE communication selector for colocated Ascend A3

miracle0517:
# Details
This PR fixes a MoE inference precision issue observed with Qwen3-30B-A3B on Ascend A3/910C in colocated RL deployments. After a live actor-to-vLLM weight update, rollout output could become incorrect from the first token, causing repetitive generation and a train_rollout_logprob_abs_diff greater than 0.1. The issue was not reproducible on 910B in colocated mode or on 910C in disaggregated mode. Weight probes confirmed that actor export, IPC transfer, expert loading, and final MoE parameter values were consistent, isolating the problem to the runtime MoE communication path. On 910C, vLLM-Ascend may select MC2, while the equivalent 910B configuration uses ALLGATHER. This change therefore forces MoECommType.ALLGATHER only when all three conditions are met: the device is A3/910C, the rollout engine is colocated, and the native selector identifies the model as MoE. Dense models, 910B devices, and disaggregated deployments retain the native communication selection. This prioritizes correctness over the MC2 performance optimization for the affected configuration.

# Testing
- [x] End-to-end Qwen3-30B-A3B colocated training on Ascend A3
image

image

### DIFF
--- a/.buildkite/npu_suites.py
+++ b/.buildkite/npu_suites.py
@@ -29,7 +29,8 @@ BUILDKITE_SOURCE = os.environ.get("BUILDKITE_SOURCE", "")
 SUITES = {
     "smk": [
         ("test_qwen3_4B_npu.py", "npu-8", "", {}),
-        ("test_qwen3_30B_A3B_npu.py", "npu-16", "", {}),
+        ("test_qwen3_30B_A3B_npu.py", "npu-8", "--colocate", {}),
+        ("test_qwen3_30B_A3B_npu.py", "npu-16", "--disaggregated", {}),
     ],
     "nightly": [],
 }
--- a/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
+++ b/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
@@ -4,7 +4,9 @@ Colocated vLLM weight sync using native IPC transfer engines.
 
 from __future__ import annotations
 
+import logging
 import os
+import sys
 from argparse import Namespace
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import asdict
@@ -25,6 +27,8 @@ from .update_weight_from_distributed import (
     post_process_weights,
     update_weights_from_distributed,
 )
+
+logger = logging.getLogger(__name__)
 
 
 class UpdateWeightFromTensor:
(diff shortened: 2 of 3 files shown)


Meihan-chen (commented):
Could we update tests/test_qwen3_30B_A3B_npu.py to colocate mode, to improve test coverage for this mode?

floatlibai:
Does this PR mean MoE colocated performance on A3 will regress? If so, roughly by how much? Is there a fix in progress, or can this be addressed on the vLLM-Ascend side so MC2 works correctly instead of falling back to ALLGATHER?

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -283,6 +287,55 @@ def _patch_npu_worker() -> None:
...
 
+    @staticmethod
+    def _patch_a3_moe_comm_selector() -> None:
+        """Force ALLGATHER for colocated MoE forwards on Ascend A3 (910C)."""
+        from vllm_ascend.utils import AscendDeviceType, get_ascend_device_type
+
+        if get_ascend_device_type() != AscendDeviceType.A3:
+            return
+
+        import vllm_ascend.ascend_forward_context as ascend_forward_context
```
momo609, line 298: Does this mean moe can only use all_gather?

miracle0517: Only in the A3 and colocate training，the patch take effect and  moe can only use allgather；in separated training，mc2 and allgather both in use。
image

miracle0517:
> Could we update tests/test_qwen3_30B_A3B_npu.py to colocate mode, to improve test coverage for this mode?

when this pr is merged, i think we can update ci test to cover colocate mode.

Meihan-chen:
> > Could we update tests/test_qwen3_30B_A3B_npu.py to colocate mode, to improve test coverage for this mode?
> 
> when this pr is merged, i think we can update ci test to cover colocate mode.

I think these two are really one step: for this fix to merge we'd want CI evidence, and CI can only produce it once the test runs in colocate mode — otherwise the patch is never exercised. So the test update is more of a prerequisite here than a follow-up. Would you mind including it in this PR? Happy to unblock the NPU gate as soon as you push.

miracle0517:
> Does this PR mean MoE colocated performance on A3 will regress? If so, roughly by how much? Is there a fix in progress, or can this be addressed on the vLLM-Ascend side so MC2 works correctly instead of falling back to ALLGATHER?

Under the scenario of Qwen3-30B-A3B, no significant performance degradation was observed.We can open an issue on vllm-ascend to see if this problem can be solved.

miracle0517:
> > > Could we update tests/test_qwen3_30B_A3B_npu.py to colocate mode, to improve test coverage for this mode?
> > 
> > 
> > when this pr is merged, i think we can update ci test to cover colocate mode.
> 
> I think these two are really one step: for this fix to merge we'd want CI evidence, and CI can only produce it once the test runs in colocate mode — otherwise the patch is never exercised. So the test update is more of a prerequisite here than a follow-up. Would you mind including it in this PR? Happy to unblock the NPU gate as soon as you push.

I did it.Can we unblock the NPU gate to test if it works correctly?

Ulrica111:
> > Does this PR mean MoE colocated performance on A3 will regress? If so, roughly by how much? Is there a fix in progress, or can this be addressed on the vLLM-Ascend side so MC2 works correctly instead of falling back to ALLGATHER?
> 
> Under the scenario of Qwen3-30B-A3B, no significant performance degradation was observed.We can open an issue on vllm-ascend to see if this problem can be solved.

We are seeing signs that forcing ALLGATHER may only be a workaround. Our results point to Vime's high-concurrency rollout and batching integration, rather than a general MC2/ALLTOALL correctness issue. Some lightweight comparison results are available in #374: https://github.com/vllm-project/vime/issues/374

.buildkite/npu_suites.py
```diff
@@ -29,7 +29,8 @@
 SUITES = {
     "smk": [
         ("test_qwen3_4B_npu.py", "npu-8", "", {}),
-        ("test_qwen3_30B_A3B_npu.py", "npu-16", "", {}),
+        ("test_qwen3_30B_A3B_npu.py", "npu-8", "--colocate", {}),
+        ("test_qwen3_30B_A3B_npu.py", "npu-16", "--disaggregated", {}),
```
Meihan-chen, line 33: Thanks for adding the mode switch. Can we keep just one 30B entry — colocate-only, still on npu-16? The 4B test already covers disaggregated in the same suite, and with a single mode we can drop the argparse flags. I think the OOM came from dropping to npu-8 rather than from colocation, so all 16 dies should give the headroom back.

miracle0517:
> > > Does this PR mean MoE colocated performance on A3 will regress? If so, roughly by how much? Is there a fix in progress, or can this be addressed on the vLLM-Ascend side so MC2 works correctly instead of falling back to ALLGATHER?
> > 
> > 
> > Under the scenario of Qwen3-30B-A3B, no significant performance degradation was observed.We can open an issue on vllm-ascend to see if this problem can be solved.
> 
> We are seeing signs that forcing ALLGATHER may only be a workaround. Our results point to Vime's high-concurrency rollout and batching integration, rather than a general MC2/ALLTOALL correctness issue. Some lightweight comparison results are available in #374: #374

Thank you for the insightful analysis. The underlying issue has been identified, and a fix is now available in the latest pull request.https://github.com/vllm-project/vime/pull/376

PR #373, opened 2026-07-27
[NPU]chore:add pip extra repo in Dockerfile.npu

yzeyu71:
### Summary
Vime depends on vllm-router. However, the Tsinghua image source used by Dockerfile.npu does not contain the vllm-router installation package. As a result, the image fails to be built. Therefore, mirrors.aliyun is added to install vllm-router.


Meihan-chen (approved):
LGTM

Issue #374, opened 2026-07-27
[NPU][A3] Garbled and repetitive MoE rollouts with inference EP after colocated IPC weight update


Ulrica111:
## Summary

We observe incorrect rollouts when running Qwen3-30B-A3B colocated RL training
on an Atlas A3 server with vLLM expert parallelism enabled.

After an actor-to-vLLM IPC weight update, some responses become garbled or
repetitive, often diverging from the first generated token.

Our current A/B results are:

| Inference EP | MoE communication | Result |
|---|---|---|
| Disabled | Default non-EP path | Correct |
| EP=8 | Native selection / MC2 | Garbled or repetitive |
| EP=8 | Forced ALLGATHER | Correct |

## Environment

- Server: Huawei Atlas A3
- NPU: 8 visible Ascend910 chips, physical IDs 0-7
- Host capacity: 16 physical chips on 8 dual-chip boards
- HBM per chip: 65536 MiB
- Model: Qwen3-30B-A3B
- Deployment: colocated RL training
- Weight synchronization: native NPU IPC
- Rollout topology: one 8-NPU engine
- Inference EP world size: 8
- Training EP size: 8
- CANN/Firmware: 9.0.0 / `9.0.0.6.200`
- `npu-smi`: `26.1.0.b070`

## Relevant configuration

```bash
python train.py \
  --train-backend megatron \
  --actor-num-nodes 1 \
  --actor-num-gpus-per-node 8 \
  --rollout-num-gpus 8 \
  --rollout-num-gpus-per-engine 8 \
  --colocate \
  \
  --hf-checkpoint /path/to/Qwen3-30B-A3B \
  --load /path/to/Qwen3-30B-A3B \
  --megatron-to-hf-mode bridge \
  --weight-update-megatron-to-hf-mode bridge \
  \
  --rollout-backend vllm \
  --vllm-weight-sync-mode native \
  --vllm-enable-expert-parallel \
  --vllm-enable-sleep-mode \
  --vllm-enforce-eager \
  --vllm-gpu-memory-utilization 0.45 \
  --vllm-max-model-len 20480 \
  \
  --rollout-batch-size 8 \
  --n-samples-per-prompt 8 \
  --global-batch-size 64 \
  \
  --tensor-model-parallel-size 4 \
  --expert-model-parallel-size 8 \
  --expert-tensor-parallel-size 1
```

## Weight-transfer verification

We inserted probes at the following stages:

1. actor-side exported weights;
[...]

PR #376, opened 2026-08-03
[NPU]fix: restore MoE ALLTOALL expert IDs after NPU memory reuse

miracle0517:
# Summary
Fix an MoE accuracy issue on Ascend A3 when running Qwen3-30B-A3B in colocated mode with expert parallelism and ALLTOALL communication.
# Problem
After a colocated weight update and NPU memory reuse, rollout inference could produce incorrect first-token log probabilities and repetitive outputs. The issue did not reproduce in disaggregated mode.
The expert_ids_per_ep_rank tensor used by the ALLTOALL token dispatcher could retain invalid contents after the vLLM sleep/wake and weight-update lifecycle. This caused received tokens to be assigned to the wrong local experts even though the model weights and ALLTOALL payload were correct.
# Fix
- Track the MoE expert-ID template generation across weight updates and wake-up operations.
- Limit the workaround to Ascend A3 colocated workers.
- Preserve existing behavior for other devices and communication paths.

# Test
- [x] End-to-end Qwen3-30B-A3B colocated training on Ascend A3
image
image

### DIFF
--- a/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
+++ b/vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
@@ -284,6 +284,47 @@ class _VLLMHijack:
         NPUWorker._npu_worker_patched = True
 
     @staticmethod
+    def _patch_a3_moe_alltoall_expert_ids() -> None:
+        """Restore the ALLTOALL expert-ID template after colocated memory reuse."""
+        from vllm_ascend.utils import AscendDeviceType, get_ascend_device_type
+
+        if get_ascend_device_type() != AscendDeviceType.A3:
+            return
+
+        from vllm_ascend.ops.fused_moe.token_dispatcher import TokenDispatcherWithAll2AllV
+
+        if getattr(TokenDispatcherWithAll2AllV, "_vime_expert_ids_patched", False):
+            return
+
+        original_dispatch_preprocess = TokenDispatcherWithAll2AllV._dispatch_preprocess
+        TokenDispatcherWithAll2AllV._vime_expert_ids_generation = 0
+
+        def _patched_dispatch_preprocess(self, hidden_states, topk_ids):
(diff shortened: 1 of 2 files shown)


vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -283,6 +283,47 @@ def _patch_npu_worker() -> None:
         _VLLMHijack._patch_one_worker(NPUWorker)
         NPUWorker._npu_worker_patched = True
 
+    @staticmethod
+    def _patch_a3_moe_alltoall_expert_ids() -> None:
```
momo609, line 287: Could this cause garbled characters in single inference? I think this issue should also be reproducible in single inference as well.

miracle0517: We noticed that a colleague has already submitted a relevant fix PR in the vllm-ascend repository. https://github.com/vllm-project/vllm-ascend/pull/13352.However, from a version control perspective, it is reasonable for us to apply a patch in vime to override the older version of vllm-ascend in order to resolve the precision issues. This approach is both justified and manageable.

CalvinXKY (approved):

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
Ulrica111: > Could this cause garbled characters in single inference? I think this issue should also be reproducible in single inference as well.

Not in our current reproduction. Single or low-concurrency requests take the MC2 path and remain normal. The issue only appears when high concurrency triggers the Vime IPC mixed prefill/decode + ALLTOALL path.

We also verified that direct vLLM-Ascend serving can use ALLTOALL normally, so this does not look like a general single-inference or ALLTOALL accuracy issue. See the lightweight comparison in [#374]

PR #382, opened 2026-08-13
[Example] Add retool tool-calling RL/SFT example (port slime → vime vLLM)

DiegoCao:
## What

Re-adds `examples/retool` — the ReTool tool-calling math RL example (SFT → RL) — ported from slime's SGLang implementation to vime's vLLM rollout path.

This is a **re-add, not a new example**: #74 deleted `retool` and `search-r1` as SGLang-only after #39's sglang→vllm rename. Adding the directory back also un-dangles the `examples/README.md` link that has pointed at nothing since then.

## The port

All the real work is the engine call in `generate_with_retool.py`:

| SGLang (upstream) | vime / vLLM |
|---|---|
| `<router>/generate` | `<router>/inference/v1/generate` |
| `{"input_ids", "sampling_params", "return_logprob": true}` | `{"model", "token_ids", "sampling_params"}` — `_build_inference_sampling_params` emits `logprobs: 1` and renames `max_new_tokens`→`max_tokens` |
| `meta_info.output_token_logprobs` | `choices[0].token_ids` + `choices[0].logprobs.content[i].logprob` |
| `meta_info.finish_reason.type` | `choices[0].finish_reason` (bare string, normalized to the nested shape) |

The choice parse is inlined as `_parse_vllm_choice` rather than imported. #178 removed the shared `_inference_generate_tokens_and_logprobs` / `_vllm_meta_from_generate_choice` helpers from `vime.rollout.vllm_rollout`, and #184 established that callers parse the choice locally — as `vllm_streaming_rollout` and `vime/agent/adapters/common.py` both do.

`tool_sandbox.py`, `requirements.txt`, `rl_data_preprocess.py` and `sft_data_processing.py` are engine-agnostic and vendored unchanged.

## Three fixes on top of a straight port

**1. Dropped `--apply-chat-template` from the RL script.** `format_conversation_with_tools` renders the whole `<|im_start|>…<|im_end|>` conversation itself, so templating in the data loader too meant the example wrapped an already-templated prompt in its own `user` turn. Rendered against Qwen3-4B's real chat template, the model received:

```
<|im_start|>user<|im_start|>user
What is 12345 * 6789?<|im_end|>
<|im_start|>assistant
<|im_end|>
<|im_start|>assistant
```

Nested `user` turns plus a spurious empty assistant turn.

**2. Take the tool-concurrency semaphore exactly once.** `tool_sandbox.SEMAPHORE` is a plain, non-reentrant `asyncio.Semaphore`, and upstream acquires it in *both* `execute_predictions` and `ToolRegistry.execute_tool` — two permits per tool call, which hangs the tool path (a single call self-deadlocks at `tool_concurrency == 1`). The limit now lives only in the registry, its natural owner.

**3. Abort instead of zero-filling missing logprobs.** `vllm_rollout.generate` substitutes `[0.0] * len(tokens)` when the engine reports no per-token logprobs. Here a missing or length-mismatched logprob array marks the sample `ABORTED` so the group returns to the buffer, rather than desyncing `rollout_log_probs` from the response tokens and silently corrupting the importance ratio.

## Tests

`tests/test_retool_generate.py` — 51 tests, mocked engine, no GPU. Covers the request body (asserting SGLang's `input_ids`/`return_logprob` did *not* survive), all five `finish_reason` mappings, the multi-turn tool loop, loss-mask/logprob alignment, context-budget clamping, observation-overflow truncation, stale-state reset on retry, prompt structure, prediction parsing, sandbox safety and reward shaping.

Registered in the existing CPU CI step, with `jinja2` added to that step's installs rather than relying on it arriving transitively via torch.

Each of the three fixes was **mutation-tested** — reintroducing the bug makes the corresponding tests fail. That mattered: the first version of the concurrency test passed even with the bug restored, because `from tool_sandbox import SEMAPHORE` binds an alias that patching `tool_sandbox.SEMAPHORE` never reaches. It was rewritten around a counting semaphore installed on both names.

## Validation

Unit tests above, plus an end-to-end smoke run on 8xH100 80GB (Qwen3-4B, GRPO,
2 rollouts x 8 prompts x 4 samples) against a **live vLLM router** — enough to
exercise the request/response contract and the tool loop, not a convergence run.

```
rollout 0: response_len/mean 844.5  rollout_time 12.9s  truncated_ratio 0.0
rollout 1: response_len/mean 777.9  rollout_time 10.7s  truncated_ratio 0.0
rollout/rollout_log_probs -0.0430   rollout/raw_reward -0.3922
grad_norm 2.2928 / 2.8861          train exit=0
```

What this confirms about the port specifically:

- The real `/inference/v1/generate` accepts the `{model, token_ids, sampling_params}`
  body and returns `token_ids` + `logprobs.content[i].logprob` in the shape
  `_parse_vllm_choice` expects.
- `rollout/rollout_log_probs = -0.043` is non-degenerate, so per-token logprobs
  genuinely flow through to training. Had the parse failed, this would be exactly
  `0.0` (the zero-fill path) or the samples would have aborted.
- The multi-turn tool loop runs end to end: the model emits a `<tool_call>`, the
  sandbox executes it, the `<interpreter>` block is appended as non-trainable
  tokens, and generation continues to `Answer: \boxed{...}`.
- `reward_func` scores the result and two optimizer steps complete.

One observation worth recording, which is upstream behaviour and deliberately
**not** changed here: the model naturally emits bare expressions
(`{"code": "64345 + 3436"}`), but `PythonSandbox` runs code as a *script* rather
than a REPL, so an expression yields no stdout and the tool returns an empty
`<interpreter>` block. The model then falls back to its own arithmetic. Making
the sandbox echo a trailing expression would change the training signal and
diverge from the published ReTool recipe, so it is left alone.

## Known gaps

- The SFT stage renders prompts with Qwen's official chat template (via `sft_rollout` + `--input-key messages`), while the RL stage uses this example's own template, which Jinja renders without the trailing newline after `<|im_start|>assistant`. That SFT/RL prompt mismatch is upstream's design and is preserved here rather than silently changed.
- `--use-kl-loss` with `--kl-loss-coef 0.00` is upstream's config and is preserved. Note `placement_group.py` creates the reference model on `kl_coef != 0 or use_kl_loss`, so the ref model is loaded and forward-passed while its KL term is multiplied by zero — dropping the flag would reclaim ~4 GB/GPU at no change to the loss.

### DIFF
--- a/examples/retool/README.md
+++ b/examples/retool/README.md
@@ -1,0 +1,135 @@
+# Retool: from SFT to RL
+
+This example demonstrates how to use the retool functionality for tool-enabled language model generation.
+
+## Overview
+
+The retool example provides:
+- Safe Python code execution in a sandbox environment
+- Tool registry for managing available tools
+- Integration with language model generation
+- Reward calculation for tool usage
+
+## Files
+
+- `generate_with_retool.py`: Main generation function with tool support
+- `tool_sandbox.py`: Tool execution and safety management
+- `sft_data_processing.py`: Process SFT dataset
+- `rl_data_preprocess.py`: Process the RL (DAPO-Math-17k) dataset
+
+## Usage
+
+1. Setup and download datasets:
+```bash
+cd vime
+pip install -e . --no-deps
+pip install -r examples/retool/requirements.txt
+# For SFT part, you can use later model to RL directly and skip SFT.
+hf download --repo-type dataset JoeYing/ReTool-SFT  --local-dir /root/JoeYing/ReTool-SFT
+hf download Qwen/Qwen3-4B-Instruct-2507 --local-dir /root/Qwen/Qwen3-4B-Instruct-2507
+
+# For RL part
+hf download --repo-type dataset zhuzilin/dapo-math-17k --local-dir /root/dapo-math-17k
+hf download --repo-type dataset zhuzilin/aime-2024  --local-dir /root/aime-2024
+# download our SFT model if you want to skip SFT
+hf download font-info/qwen3-4b-sft-SGLang-RL --local-dir /root/font-info/qwen3-4b-sft
+```
+
+2. Create torch dist
+
+Both checkpoints use rope theta `5e6`, which differs from the `1e6` default in
+`scripts/models/qwen3-4B.sh`. Override it with `MODEL_ARGS_ROTARY_BASE` so the
+conversion and the training scripts agree.
+
+For SFT
+```bash
+MODEL_ARGS_ROTARY_BASE=5000000 source scripts/models/qwen3-4B.sh
+PYTHONPATH=/root/Megatron-LM python tools/convert_hf_to_torch_dist.py \
+    ${MODEL_ARGS[@]} \
+    --hf-checkpoint /root/Qwen/Qwen3-4B-Instruct-2507 \
+    --save /root/Qwen/Qwen3-4B-Instruct-2507_torch_dist
+```
+
+Or RL only
+```bash
+MODEL_ARGS_ROTARY_BASE=5000000 source scripts/models/qwen3-4B.sh
+PYTHONPATH=/root/Megatron-LM python tools/convert_hf_to_torch_dist.py \
+    ${MODEL_ARGS[@]} \
+    --hf-checkpoint /root/font-info/qwen3-4b-sft \
+    --save /root/font-info/qwen3-4b-sft_torch_dist
+```
+
+3. SFT:
+```bash
+python examples/retool/sft_data_processing.py
+bash examples/retool/retool_qwen3_4b_sft.sh
+```
+
+4. RL:
+```bash
+bash examples/retool/retool_qwen3_4b_rl.sh
+```
+
+5. Use in your training scripts by importing the generate function:
+```python
+from generate_with_retool import generate, reward_func
+```
+
+The RL script wires these up with:
+```bash
+--custom-generate-function-path generate_with_retool.generate
+--custom-rm-path generate_with_retool.reward_func
+```
+`generate_with_retool` is resolved as a top-level module, so the example
+directory is added to `PYTHONPATH` in the script's Ray runtime env (this is also
+what lets it import its sibling `tool_sandbox`).
+
+## Tool Format
+
+The system uses the following tool format:
+
+```
+You may call one or more functions to assist with the user query.
+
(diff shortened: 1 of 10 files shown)


CalvinXKY:
For an example merge, the 2-rollout smoke is not enough, it only checks the vLLM contract and that one train step runs.
Please post a longer RL run with WandB (or equivalent) showing reward/eval converging (AIME is already wired at --eval-interval 20). Without that curve, we cannot tell whether the ported recipe actually trains.

DiegoCao:
@CalvinXKY Yes, we can add test trajectory beyond that.

tests/test_retool_generate.py
```diff
@@ -0,0 +1,572 @@
...
+    def __call__(self, text: str, add_special_tokens: bool = False):
+        assert add_special_tokens is False
+        return {"input_ids": [ord(c) for c in text]}
+
+    def decode(self, token_ids, skip_special_tokens: bool = True) -> str:
+        return "".join(chr(int(t)) for t in token_ids)
+
+
+class _FakeState:
+    def __init__(self, args):
```
aoshen02, line 63: We don't need a test for it.

examples/retool/generate_with_retool.py
```diff
@@ -0,0 +1,485 @@
+# Adapted from https://github.com/volcengine/verl/blob/cb809d66e46dfd3342d008628891a14a054fa424/recipe/retool/retool.py
+# Ported from slime's SGLang-based example to vime's vLLM ``/inference/v1/generate`` path.
```
aoshen02, line 2: No need to add these, we port most of the code from slime

examples/retool/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+Note that `PythonSandbox._check_code_safety` is deliberately strict: it allows
+only the stdlib modules in `PythonSandbox.allowed_modules` (`math`, `random`,
+`statistics`, `decimal`, `fractions`, …) and rejects `eval`/`exec`/`open`,
+dunder access, and imports outside that set. Widen `allowed_modules` if your task
+needs more (e.g. `sympy` or `numpy`).
+
+## Notes on the vLLM port
+
+This example was ported from slime's SGLang implementation. The rollout loop
```
aoshen02, line 120: And here as well. It is very clear that vime is a fork of slime

PR #385, opened 2026-08-14
Apply-MegatronAdaptor-NPU-migration-to-clean-branch

momo609:
## Summary

- Remove MindSpeed-specific dependencies and imports.
- Switch NPU backend integration to MegatronAdaptor and TransformerEngineNPU.
- Align Megatron-related patches with the mainline implementation.
- Remove obsolete MindSpeed and Megatron communication patches.
- Update Megatron, vLLM, and bridge patch application order.
- Fix MegatronAdaptor initialization and feature re-patching in the actor.
- Update NPU attention patch and weight-transfer compatibility logic.
- Fix nested NPU memory-pool allocation during training offload.
- Update the related unit test and deployment scripts.

## Files Changed

- `vime/backends/megatron_utils/`
- `vime/utils/`
- `docker/`
- `scripts/`
- `examples/`
- `tests/`
- `train.py`
- `train_async.py`

PR #390, opened 2026-08-19
[NPU][CI] Refactor patch reconciliation

Meihan-chen:
## Summary

- replace the hard-coded NPU patch inventory with an ordered `series.conf`
- persist the image-applied patch series and exact patch bytes under `/opt/npu_patch`
- reconcile patch changes through deterministic digest comparison, reverse-order revert, and forward-order apply
- support patch additions, deletions, renames, reordering, and nested source paths without patch-specific CI branches
- document when the default image can be reused and when `image-build` is required
- add an 8-NPU colocated Qwen3-VL-8B GRPO smoke test to the `smk` suite
- fix missing MindSpeed configuration fields when ModelOpt creates transformer configs before MindSpeed patches are registered

## Motivation

The previous updater used the current checkout's static `PATCH_CONFIGS` to represent both the image's OLD patch state and the requested NEW state. This is not reliable when patches are deleted, renamed, reordered, or modified.

PR #385 exposed this issue in [NPU CI build 167](https://buildkite.com/vllm/vime-npu-ci/builds/167).
The updater did not retain the deleted `megatron_comm.patch` and `mindspeed.patch`, failed to revert the image's Megatron patch, and then tried to apply the new patch on top of the remaining OLD state.

## Design

The image stores its ordered OLD series and exact patch bytes under `/opt/npu_patch`. The current Vime checkout provides the NEW series through `docker/npu_patch/series.conf`.

Each entry uses:

```text
target_worktree|image_patch|source_patch
```
If the OLD and NEW digests differ, CI reverts the OLD series from bottom to top and applies the NEW series from top to bottom. This allows the image to revert patches that have already been removed from the current checkout.

## Contributor impact

Patch changes remain explicit and reviewable. Contributors must update the patch file, the corresponding `docker/Dockerfile.npu` apply operation, and `docker/npu_patch/series.conf` in the same change. Ordinary patch lifecycle changes no longer require patch-specific CI logic.

### DIFF
--- a/.buildkite/README-NPU-CI.md
+++ b/.buildkite/README-NPU-CI.md
@@ -86,7 +86,22 @@ list in `SUITES`.
 
 ## Adding or removing a patch
 
-Patches live in `docker/npu_patch/` and are registered in the `PATCH_CONFIGS`
-dict in [`update-npu-environment.sh`](./scripts/update-npu-environment.sh).
-Update both when adding or removing a patch.
+Patches are applied explicitly by [`docker/Dockerfile.npu`](../docker/Dockerfile.npu)
+and reconciled at runtime in the order declared by
+[`docker/npu_patch/series.conf`](../docker/npu_patch/series.conf).
 
+When adding a patch file, add the matching Dockerfile apply operation and the
+`series.conf` entry in the same order. When deleting a patch, remove the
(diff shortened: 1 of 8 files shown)


CalvinXKY (approved):
LGTM

PR #398, opened 2026-08-23
docker: split pull-weight patch and drop upstreamed vLLM fixes

aoshen02:
## Summary

- split delta checkpoint reconstruction and the worker `pull_weights` entrypoint into the standalone image patch `docker/patch/latest/vllm-pull_weights.patch`
- apply the pull-weight patch before the general vLLM integration patch in the Docker image
- remove the FlashInfer NVL token-cap workaround now covered by vLLM #45989/current nightly behavior
- remove Vime's duplicate core per-request speculative decoding accounting now provided by vLLM #48915
- keep the small TITO response adapter and enable native `per_request_spec_decode_metrics=summary` for speculative rollout

## Patch boundaries

`vllm-pull_weights.patch` contains only:

1. host-local full/delta checkpoint reconstruction
2. `Worker.pull_weights`
3. focused local-checkpoint unit tests

`vllm.patch` retains only integrations that are not yet native in the current nightly:

1. rank-local IPC payload routing (vLLM #52497 is still open)
2. TITO weight-version and speculative-metrics response fields

## Validation

Validated against `vllm/vllm-openai:nightly` source commit `e9d1398d9edfd90fcc1cf783805240e3effec013`:

- both patches apply sequentially with plain `git apply`
- `tests/model_executor/model_loader/test_local_checkpoint.py`: 4 passed
- Vime focused tests: 99 passed
- pre-commit: passed
- `git diff --check`: passed

### H200 E2E

All tests used `vllm/vime:latest` at Vime commit `052693a89b2e7cbbda3de576f41ea609fb0b74a1`:

- `tests/test_qwen3_4B_external_pd.py`, delta mode, 6 GPUs: passed
- `tests/test_full_disk_weight_update.py`, 4 GPUs: passed
- `tests/test_mimo_7B_mtp_only_grad.py`, 8 GPUs: passed
- `tests/test_moonlight_16B_A3B.py`, 8 GPUs: passed
- Moonlight rank-local colocated weight update: 4.5 s

The Qwen3-4B delta run observed:

- steady updates: 8.32 s and 11.0 s, mean 9.66 s
- changed-byte density: 0.62%
- wire payload: 0.14 GB

### Slime comparison

For the same Qwen3-4B BF16 topology (4 trainer GPUs + 1 prefill GPU + 1 decode GPU), the prior Slime crossed-host A/B measured:

- full-disk steady mean: 7.9 s
- delta-disk steady mean: 7.0 s
- first real delta update: 21.8 s
- delta payload: 0.13-0.17 GB at 0.61-0.80% density

The current Vime single-host sample has matching density and payload, but its 9.66 s steady mean is about 38% slower than the Slime A/B mean. This is directional only because the current Vime result has two steady samples and was not repeated with a host crossover.

A separate strict two-node Qwen3-30B-A3B comparison measured:

- Vime delta-disk nonzero update: 39.414 s
- Slime delta-disk nonzero update: 47.627 s
- Vime was 17.25% faster in that case
- Vime NCCL median remained much faster at 17.221 s, showing that large-model delta is still dominated by CPU diff/apply work

## Published image

`vllm/vime:latest` is published as a multi-architecture image:

- digest: `sha256:0225926e4a78de737377aca20ee0a841bc87bce2ccaf83964bb47d1b1a275319`
- platforms: `linux/amd64`, `linux/arm64`

The previously published base image used vLLM `311b3513a`, while the repository patch targets current nightly `e9d1398`. The incremental image therefore retains the legacy spec-metrics implementation for that old base; Vime accepts both the legacy and native current-nightly response field names.


aoshen02:
## Image and E2E validation

Published `vllm/vime:latest`:

- Multi-arch digest: `sha256:0225926e4a78de737377aca20ee0a841bc87bce2ccaf83964bb47d1b1a275319`
- Platforms: `linux/amd64`, `linux/arm64`
- Vime commit: `052693a89b2e7cbbda3de576f41ea609fb0b74a1`

E2E results on H200:

- `tests/test_qwen3_4B_external_pd.py` with `VIME_TEST_UPDATE_MODE=delta`, 6 GPUs: **passed**
  - steady delta update observed: density `0.62%`, wire payload `0.14 GB`
- `tests/test_full_disk_weight_update.py`, 4 GPUs: **passed**
- `tests/test_mimo_7B_mtp_only_grad.py`, 8 GPUs: **passed**
- `tests/test_moonlight_16B_A3B.py`, 8 GPUs: **passed**
  - rank-local colocated weight update completed in `4.5s`

The previously published image still used vLLM `311b3513a`, while the repository patch in this PR targets current nightly `e9d1398`. The incremental image therefore uses the legacy spec-metrics implementation for that old base; Vime now accepts both legacy and native current-nightly metric field names. The repository patch remains cleanly validated against current nightly.

Moonlight's 30 GB converted checkpoint was placed on `/mnt/ephemeral` to avoid exhausting the host root filesystem; this did not change the test configuration or code path.

princepride (approved):

PR #400, opened 2026-08-26
[CI/Build] Run the CPU tests CI silently skips, and guard against new gaps

natedemoss:
## Problem

CI enumerates test files by hand — one `python tests/<file>.py` line per test in `.buildkite/pipeline.yml`, one `(file, num_gpus, ...)` tuple in `.buildkite/gpu_suites.py`. Nothing walks `tests/`, so a test only runs if its author also wired it into a job. Two failure modes, both of which leave CI green:

1. **The file is never wired in.** Three existing files are in this state today and have never run in CI:
   - `tests/test_qwen3_linear_attention_cu_seqlens.py`
   - `tests/test_chunked_gae.py`
   - `tests/test_rollout_metrics.py`

2. **The CPU job's invocation style fails open.** `python tests/<file>.py` on a module with no `if __name__ == "__main__"` block just imports it: zero `test_*` functions execute and the command exits 0 without asserting anything.

## Changes

- **Wire the unreferenced files into the right jobs.** `test_qwen3_linear_attention_cu_seqlens.py`, `test_chunked_gae.py`, and the new `test_ci_test_coverage.py` go into *plugin contracts & CPU tests* (all three are CPU-safe). `test_rollout_metrics.py` goes into *synchronized upstream CPU tests*, which runs in `vllm/vime:latest` — it imports `vime.ray.rollout`, so it needs vLLM present.
- **Add `__main__` entry points** to the two files the CPU job invokes as bare `python <file>`, matching every other file in that job.
- **Add `tests/test_ci_test_coverage.py`** so neither gap can reopen. It fails when a test file is neither referenced by a CI job nor waived in `NOT_RUN_IN_CI`, and when a file CI runs as bare `python <file>` has no `__main__` entry point. The waiver list is itself checked for staleness (entries must name a file that still exists and that CI still does not run), so it cannot rot.

`NOT_RUN_IN_CI` starts with five entries, each with the reason: the two GLM-5.2 e2e tests whose `run_gate()` is disabled at the source, and three multi-GPU end-to-end runs no `gpu_suites` entry selects.

## Verification

- `pytest tests/test_ci_test_coverage.py` — 4 passed.
- Confirmed against `main` that the three existing files had zero references anywhere in `.buildkite/`, and that dropping any part of the pipeline change makes the new test fail with the offending file named.
- `ruff check`, `black --check` (24.3.0), and `isort --profile=black` (5.13.2) clean on all touched files; `pipeline.yml` parses.
- `test_chunked_gae.py` takes ~3 min on CPU (noted in a comment in the pipeline) — the 256x128K cases are dominated by the serial `vanilla_gae` reference the parallel scan is checked against. The other additions are fast.

## AI assistance

This PR was written with AI assistance (Claude). I reviewed every changed line, ran the tests above locally, and verified the coverage claims against `main` myself. The commit carries a `Co-authored-by:` trailer.


aoshen02:
Thanks for the pr, test_discounted_returns.py will be added by pr 402 and so does the changes in tests/test_chunked_gae.py and tests/test_qwen3_linear_attention_cu_seqlens.py. We don' consider test_ci_test_coverage.py for now.

CalvinXKY (approved):

PR #403, opened 2026-08-28
feat: support GLM-4.7-Flash (MLA + MoE) on Ascend NPU

is-not:
## Overview

Adds training/inference support for the GLM-4.7-Flash model on Ascend 910B NPU. GLM-4.7-Flash uses DeepSeek-V3-style Multi-Head Latent Attention (MLA) plus a GLM-style MoE (64 routed experts + 1 shared expert, sigmoid router with expert bias), and supports Multi-Token Prediction (MTP). This PR registers the corresponding megatron bridge, adds the model-argument and run scripts, and fixes an MTP drafter compatibility issue under NPU cudagraph capture.

### New bridge plugin
- **`vime_plugins/megatron_bridge/glm4moe_lite.py`**: Adds the megatron bridge for `Glm4MoeLiteForCausalLM`. It subclasses `GLM45Bridge` to reuse its MTP plumbing (`num_nextn_predict_layers` mapping loop, fused-expert handling, etc.) and only overrides `provider_bridge` and `mapping_registry` for the MLA model. MindSpeed TE module types are registered via the public `AutoMapping.register_module_type` API through `_register_mindspeed_te_module_types()`, so weight loading works on NPU without manually patching the container's `param_mapping.py`. Non-MLA branches that are always dead code for this model (`is_mla` / `is_mla_mapping` are always true) have been removed, along with the now-unused `QKVMapping` / `GPTModelProvider` imports.
- **`vime_plugins/megatron_bridge/__init__.py`**: Imports and registers the new bridge.

### MTP cudagraph compatibility fix
- **`docker/npu_patch/vllm.patch`**: Adds a static patch to vLLM's upstream Glm4MoeLiteMultiTokenPredictorLayer.forward that rewrites the bool-mask index assignment (inputs_embeds[positions == 0] = 0) into the equivalent element-wise torch.where. The original form routes to aclnnNonzeroV2 (a data-dependent-shape op), which fails with stream is captured under NPU cudagraph capture. After the rewrite, the MTP drafter runs under cudagraph on Ascend. The patch is applied at image build time (reconciled at runtime via series.conf), so no runtime hook or env gate is needed. The change is scoped to the GLM MTP class and is a no-op for non-GLM models.

### Run scripts
- **`scripts/models/glm4.7-30B-A3B-npu.sh`**: Model argument definitions for GLM-4.7-Flash 30B-A3B (MoE config, MLA q/kv lora rank, qk/v head dim, etc.).
- **`scripts/run-glm4.7-30B-A3B-npu.sh`**: NPU training entry script (GRPO, 16 cards on a single node); enables MTP training and vLLM MTP speculative decoding, and sets `VIME_PATCH_GLM_MTP_GRAPH=1`.

## Usage

```bash
bash scripts/run-glm4.7-30B-A3B-npu.sh
```

Weight and data paths can be overridden via the `DATA_ROOT` environment variable.

## Testing

- GLM-4.7-Flash GRPO training runs through on a single NPU node with 16 cards; weight loading is normal and the reward keeps trending upward.

mtp-spec1-patch

- Whether to enable MTP or not

mtp-spec1-patch-vs-nomtp

### DIFF
--- a/scripts/run-glm4.7-30B-A3B-npu.sh
+++ b/scripts/run-glm4.7-30B-A3B-npu.sh
@@ -1,0 +1,155 @@
+#!/bin/bash
+
+# for rerun the task
+pkill -9 -f '[v]llm serve|VLL[M]::'
+pkill -9 -f VLLM
+sleep 3
+ray stop --force
+pkill -9 ray
+pkill -9 python
+sleep 3
+pkill -9 ray
+pkill -9 python
+pkill -9 redis
+
+set -ex
+
+export PYTHONUNBUFFERED=1
+export ASCEND_RT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15
+export RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES=1
+export CUDA_DEVICE_MAX_CONNECTIONS=1
+export HCCL_HOST_SOCKET_PORT_RANGE=60000-60050
+export HCCL_NPU_SOCKET_PORT_RANGE=61000-61050
+export HYDRA_FULL_ERROR=1
+export DISABLE_L2_CACHE=1
+export VLLM_ASCEND_ENABLE_NZ=0
+export VLLM_USE_AOT_COMPILE=0
+export PYTHONPATH="/root/Megatron-Bridge/src:/root/Megatron-LM/:${PYTHONPATH:-}"
+
+# Apply the vLLMWorkerExtension monkey-patch that makes the GLM-4.7 MTP drafter
+# forward cudagraph-friendly on NPU (torch.where instead of bool-mask index,
+# which otherwise fails aclnnNonzeroV2 under cudagraph capture).
+# Drop this once vLLM upstream fixes the graph-safety issue.
+export VIME_PATCH_GLM_MTP_GRAPH=1
+
+unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY
+
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+source "${SCRIPT_DIR}/models/glm4.7-30B-A3B-npu.sh"
+
+DATA_ROOT="${DATA_ROOT:-/root}"
+
+CKPT_ARGS=(
+   --hf-checkpoint ${DATA_ROOT}/weights/GLM-4.7-Flash/
+   --load ${DATA_ROOT}/weights/GLM-4.7-Flash/
+   --ref-load ${DATA_ROOT}/weights/GLM-4.7-Flash/
+   --megatron-to-hf-mode bridge
+)
+
+ROLLOUT_ARGS=(
+   --prompt-data ${DATA_ROOT}/datasets/dapo-math-17k/dapo-math-17k.jsonl
+   --input-key prompt
+   --label-key label
+   --apply-chat-template
+   --rollout-shuffle
+   --rm-type deepscaler
+   --num-rollout 3000
+   --rollout-batch-size 32
+   --n-samples-per-prompt 8
+   --rollout-max-response-len 8192
+   --rollout-temperature 1
+   --global-batch-size 256
+   --balance-data
+)
+
+EVAL_ARGS=(
+   --eval-interval 20
+   --eval-prompt-data aime ${DATA_ROOT}/datasets/aime-2024/aime-2024.jsonl
+   --n-samples-per-eval-prompt 16
+   --eval-max-response-len 16384
+   --eval-top-p 1
+)
+
+PERF_ARGS=(
+   --tensor-model-parallel-size 4
+   --sequence-parallel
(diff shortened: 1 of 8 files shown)


scripts/run-glm4.7-30B-A3B-npu.sh
```diff
@@ -0,0 +1,155 @@
...
+   --attention-softmax-in-fp32
+   --attention-backend flash
+
+   --use-flash-attn
+   --no-gradient-accumulation-fusion
+)
+
+# launch the master node of ray in container
+export MASTER_ADDR=${MASTER_ADDR:-"127.0.0.1"}
+ray start --head --node-ip-address ${MASTER_ADDR} --num-gpus 16 --disable-usage-stats --dashboard-host=0.0.0.0 --dashboard-port=8265
```
CalvinXKY: `ray start --num-gpus 16` advertises GPU capacity on an NPU job. On Ascend, vime schedules NPU bundles (see `launch.py`: `--num-gpus 0 --resources '{"NPU": 16}'`). Please align with the existing NPU launch convention (cf. `run-qwen3-30B-A3B-npu.sh` / `launch.py`).

CalvinXKY:
DCO is failing: both commits are missing `Signed-off-by:` lines. Please add sign-off before merge.

vime_plugins/megatron_bridge/glm4moe_lite.py
```diff
@@ -0,0 +1,376 @@
...
+        provider.gated_linear_unit = True
+        provider.add_bias_linear = False
+        provider.share_embeddings_and_output_weights = False
+        provider.multi_latent_attention = True
+        provider.qk_layernorm = True
+
+        provider.moe_shared_expert_overlap = True
+        provider.moe_token_dispatcher_type = "alltoall"
+        provider.moe_router_load_balancing_type = "seq_aux_loss"
+        provider.moe_router_pre_softmax = False
```
CalvinXKY: provider.moe_router_pre_softmax = False conflicts with --moe-router-pre-softmax in the model script and with NVIDIA's GLM47FlashBridge (True). Please align provider defaults with the training CLI / official bridge to avoid router behavior mismatch during HF bridge load.

Meihan-chen (commented):
Please rebase onto the latest ascend branch and add a new GLM test to validate the integration and prevent regressions，the new end-to-end GLM path is not covered by NPU CI.

vime/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -416,6 +416,55 @@ def _npu_rotary_emb_init(
         ApplyRotaryEmb.__init__ = _npu_rotary_emb_init  # type: ignore[attr-defined]
         ApplyRotaryEmb._npu_rotary_patched = True
 
+    @staticmethod
+    def _patch_glm_mtp_graph_friendly() -> None:
+        """Patch GLM-4.7 MTP draft forward to be NPU cudagraph-friendly.
+
```
momo609: this patch use for mtp support aclgraph?

scripts/run-glm4.7-30B-A3B-npu.sh
```diff
@@ -0,0 +1,149 @@
...
+   --overlap-cpu-optimizer-d2h-h2d
+   --use-precision-aware-optimizer
+)
+
+
+VLLM_ARGS=(
+   --rollout-num-gpus-per-engine 4
+   --vllm-gpu-memory-utilization 0.7
+   --vllm-cudagraph-capture-sizes 1 2 4 8 $(seq 16 8 256)
+   --vllm-speculative-config '{"method":"mtp","num_speculative_tokens":1}'
```
Meihan-chen, line 117: Just curious, what’s the benefit of enabling MTP here? If it’s optional, maybe we could get the model support in first and add MTP for a follow-up?

Meihan-chen: With the main sync in #413 changing the model-loading and weight-sync paths quite a bit, how about focusing on basic GLM-4.7-Flash support here and adding MTP after the sync lands? That should help avoid having to adapt and validate the MTP path twice.

is-not: From our testing, the benefits and adaptation cost of MTP are quite clear: on the inference side, with spec=1 configuration, rollout end-to-end latency is reduced by ~12%, the pos0 acceptance rate reaches ~78%, and the gains from speculative decoding are clearly realized. On the training side, MTP loss aligns perfectly with the convergence pace of the main task; the reward curve overlaps with the non-MTP baseline (both converge from 0.52 to 0.73), so MTP does not degrade training performance. In terms of adaptation cost, training and inference share the same set of checkpoints with no extra weight files required. Code changes are confined to the mapping_registry in the bridge plugin, adding only about 60 lines of MTP layer mapping logic with zero intrusion into the main framework path. MTP is fully optional: when disabled, simply remove MTP_ARGS and vllm-speculative-config related settings, and the model will run normally and complete RL convergence. All things considered, I recommend including MTP in this PR.

Meihan-chen (approved):

PR #405, opened 2026-08-31
ci(rocm): rename ROCm queue to amd_mi355_vime_rl

indianspeedster:
Renames the self-hosted ROCm Buildkite queue from `amd_gfx950` to `amd_mi355_vime_rl` in `.buildkite/pipeline-rocm.yaml` (all three GPU steps + the header comment).

`amd_gfx950` appeared nowhere else in the repo. The remaining `gfx950` references (`docker/Dockerfile.rocm`, `scripts/run-qwen3-8B*-amd.sh`) are the GPU arch, not the queue, and are unchanged.

**Requires first:** the `amd_mi355_vime_rl` queue must exist in the Buildkite cluster and the MI355 agents must advertise the matching `queue` tag — otherwise these steps have no matching agent. They sit behind the manual block gate, so nothing else is affected.

### DIFF
--- a/.buildkite/pipeline-rocm.yaml
+++ b/.buildkite/pipeline-rocm.yaml
@@ -1,9 +1,9 @@
 # Buildkite CI for vime — AMD ROCm GPU suites. See .buildkite/README.md.
 #
-# One step per test on the self-hosted gfx950 (MI350X) queue (queue=amd_gfx950,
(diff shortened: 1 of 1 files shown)


natedemoss:
lgtm

aoshen02 (approved):

PR #406, opened 2026-09-01
[Bugfix] Restore DSpark draft param views after TMS resume

CalvinXKY:
## Summary

- **Restore DSpark draft param views after TMS resume** (commit `8fbfd891`): Fixes param view corruption when TorchMemorySaver pause/resume interacts with DSpark draft model weight sync.
- **Patch vLLM DFlash/DSpark meta-device bug** (commit `934a91ff`): Runtime monkeypatch for [vllm-project/vllm#55076](https://github.com/vllm-project/vllm/pull/55076). IPC engine weight sync leaves some draft attention params on meta device; `_build_context_kv_buffers` then crashes on `torch.cat` with mixed CUDA/meta tensors. The patch skips rebuild when previous CUDA buffers exist (incremental IPC update), raises on first-load failures, and recomputes RoPE `cos_sin_cache` from meta.

## Dependencies

- Requires vLLM with IPC engine support (vLLM 0.27.2+)
- vLLM meta-device fix: [vllm-project/vllm#55076](https://github.com/vllm-project/vllm/pull/55076) (temporary runtime patch included; will be removed after upstream merge)

## What was removed

- DSpark smoke test (`tests/test_qwen3_4B_dspark_short.py`) — blocked by TE 2.16.1 sm80 incompatibility on A800; will re-add after vLLM upstream fix lands
- CI assertions for `spec_accept_rate` in `train_metric_utils.py`

## Test Plan

- [x] DSpark 4B training verified on A800 (non-colocate, 100 steps)
- [x] vLLM meta-device patch verified: 5/5 unit tests pass on .62
- [ ] vLLM upstream PR #55076 merged
- [ ] Remove runtime patch after upstream fix is in base image
- [ ] Re-add DSpark smoke test

### DIFF
--- a/vime/backends/megatron_utils/actor.py
+++ b/vime/backends/megatron_utils/actor.py
@@ -196,6 +196,12 @@ class MegatronTrainRayActor(TrainRayActor):
 
         torch_memory_saver.resume()
 
+        # Restore DSpark draft param views after TMS resume
+        if getattr(self.args, "dspark_enabled", False):
+            from vime.backends.megatron_utils.dspark.modeling import restore_dspark_param_views
+
+            restore_dspark_param_views(self.model)
+
         clear_memory()
         reload_process_groups()
 
--- a/vime/backends/megatron_utils/dspark/modeling.py
+++ b/vime/backends/megatron_utils/dspark/modeling.py
@@ -600,3 +600,35 @@ def attach_dspark_model(model, args, config) -> None:
     )
     for param in model.draft_model.parameters():
         param.grad_norm_group = "dspark"
+
+
+def restore_dspark_param_views(model_chunks):
+    """Restore view relationship between DSpark draft params and DDP buffer after TMS resume.
+
+    After torch_memory_saver.resume(), param.data tensors are no longer views into
+    the DDP contiguous buffer. This function rebinds each draft parameter to its
+    slice in the DDP buffer, so subsequent optimizer.step() updates are visible.
+
+    This should be called once after TMS resume, not after every optimizer step.
+    """
+    from megatron.core.utils import unwrap_model
+
+    for chunk in model_chunks:
+        if not hasattr(chunk, "buffers"):
+            continue
+        unwrapped = unwrap_model(chunk)
+        draft = getattr(unwrapped, "draft_model", None)
+        if draft is None:
+            continue
+        draft_param_ids = {id(p) for p in draft.parameters()}
+        for buffer in chunk.buffers:
+            pim = buffer.param_index_map
+            for param_obj, (_start, _end, bucket_id) in pim.items():
+                if id(param_obj) not in draft_param_ids:
+                    continue
+                bucket = buffer.buckets[bucket_id]
+                if hasattr(bucket, "param_to_index") and param_obj in bucket.param_to_index:
+                    local_start, local_end = bucket.param_to_index[param_obj]
+                    if isinstance(local_start, int):
+                        # Restore view: rebind param.data to DDP buffer slice
+                        param_obj.data = bucket.param_data.view(-1)[local_start:local_end].view(param_obj.data.shape)
--- a/vime/backends/vllm_utils/dspark_patch.py
+++ b/vime/backends/vllm_utils/dspark_patch.py
@@ -1,0 +1,115 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Runtime patch for vLLM DFlash/DSpark meta-device tensor bug.
+
+Applies the fix from vllm-project/vllm#55076 at import time so that
+DSpark/DFlash draft models survive IPC engine weight sync without crashing
+on meta-device tensors in ``_build_context_kv_buffers``.
+
+Remove this module once vLLM #55076 is merged and the base image includes it.
+"""
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+_patched = False
+
+
+def patch_vllm_dspark_meta_device():
+    """Monkeypatch ``qwen3_dflash._build_context_kv_buffers`` and
+    ``_build_fused_kv_buffers`` to materialize meta-device tensors to CUDA
+    before concatenation.
+
+    This is a temporary workaround for vLLM versions that do not yet include
+    the upstream fix (vllm-project/vllm#55076).
+    """
+    global _patched
+    if _patched:
+        return
+    _patched = True
+
+    try:
+        import torch
+
+        from vllm.model_executor.models import qwen3_dflash as mod
+
+        def _is_meta(t):
+            return t is not None and t.device.type == "meta"
+
+        def _resolve_device(*tensors):
+            for t in tensors:
+                if t is not None and t.device.type != "meta":
+                    return t.device
+            return torch.device("cuda", torch.cuda.current_device())
+
+        def _build_context_kv_buffers_patched(self, layers_attn, has_bias):
+            self._hidden_norm_weight = self.hidden_norm.weight.data
+
+            kv_weights = [a.qkv_proj.weight[a.q_size :] for a in layers_attn]
+            kv_biases = [a.qkv_proj.bias[a.q_size :] for a in layers_attn] if has_bias else []
+            k_norm_weights = [a.k_norm.weight.data for a in layers_attn]
+            needed = [self._hidden_norm_weight, *kv_weights, *k_norm_weights, *kv_biases]
+
+            if any(_is_meta(t) for t in needed):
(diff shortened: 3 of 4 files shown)


CalvinXKY:
# Tests

```
#!/bin/bash

# DSpark 4-GPU 100-step: DSpark ON, policy NOT frozen
# Goal: Fair comparison with baseline - main model trains via GRPO, draft trains via 3-loss
# GPUs: 0-3 | TP=2 | colocate | 4 GPUs
#
# Key change: --dspark-freeze-policy REMOVED (policy model trains normally)
# DSpark config:
#   - Pretrained draft model (Qwen3-4B-dspark-pretrained)
#   - 5 draft layers, block_size=7, num_speculative_tokens=7
#   - 3-loss training: CE + L1 + confidence
#   - Policy model trains via GRPO policy_loss (NOT frozen)

set -ex

# Cleanup residual processes
pkill -9 -f '[v]llm serve|VLL[M]::' || true
sleep 3
ray stop --force || true
pkill -9 ray || true
pkill -9 python || true
sleep 3
pkill -9 ray || true
pkill -9 python || true

export PYTHONUNBUFFERED=1

# Detect NVLink
NVLINK_COUNT=$(nvidia-smi topo -m 2>/dev/null | grep -o 'NV[0-9][0-9]*' | wc -l)
if [ "$NVLINK_COUNT" -gt 0 ]; then
    HAS_NVLINK=1
else
    HAS_NVLINK=0
fi
echo "HAS_NVLINK: $HAS_NVLINK (detected $NVLINK_COUNT NVLink references)"

NUM_GPUS=4
echo "NUM_GPUS: $NUM_GPUS"

# Model config (Qwen3-4B) - inlined from models/qwen3-4B.sh
MODEL_ARGS=(
   --swiglu
   --num-layers 36
   --hidden-size 2560
   --ffn-hidden-size 9728
   --num-attention-heads 32
   --group-query-attention
   --num-query-groups 8
   --use-rotary-position-embeddings
   --disable-bias-linear
   --normalization "RMSNorm"
   --norm-epsilon 1e-6
   --rotary-base 1000000
   --vocab-size 151936
   --kv-channels 128
   --qk-layernorm
)

CKPT_ARGS=(
   --hf-checkpoint /data/nfs_87/xky/models/Qwen3-4B
   --ref-load /data/nfs_87/xky/models/Qwen3-4B_torch_dist
   --save /data/nfs_87/xky/dspark_test/save_dspark_4gpu_100step_nofreeze
   --save-interval 50
)

ROLLOUT_ARGS=(
   --prompt-data /data/nfs_87/datasets/dapo-math-17k/dapo-math-17k.jsonl
   --input-key prompt
   --label-key label
   --apply-chat-template
   --rollout-shuffle
   --rm-type deepscaler
   --num-rollout 100
   --rollout-batch-size 32
   --n-samples-per-prompt 8
   --rollout-max-response-len 4096
   --rollout-temperature 1

   --global-batch-size 256
   --balance-data
)

EVAL_ARGS=(
   --eval-interval 20
   --eval-prompt-data aime /data/nfs_87/datasets/aime-2024/aime-2024.jsonl
   --n-samples-per-eval-prompt 16
   --eval-max-response-len 16384
   --eval-top-p 1
)

PERF_ARGS=(
   --sequence-parallel
   --context-parallel-size 1
   --expert-model-parallel-size 1
   --expert-tensor-parallel-size 1

   --recompute-granularity full
   --recompute-method uniform
   --recompute-num-layers 1

   --use-dynamic-batch-size
   --max-tokens-per-gpu 9216
)

GRPO_ARGS=(
   --advantage-estimator grpo
   --use-kl-loss
   --kl-loss-coef 0.00
   --kl-loss-type low_var_kl
   --entropy-coef 0.00
   --eps-clip 0.2
   --eps-clip-high 0.28
)

OPTIMIZER_ARGS=(
   --optimizer adam
   --lr 1e-6
   --lr-decay-style constant
   --weight-decay 0.1
   --adam-beta1 0.9
   --adam-beta2 0.98
)

VLLM_ARGS=(
   --rollout-num-gpus-per-engine 2
   --vllm-gpu-memory-utilization 0.5
   --vllm-enforce-eager
   --vllm-speculative-config '{"method":"dspark","model":"/data/nfs_87/xky/models/Qwen3-4B-dspark-pretrained","num_speculative_tokens":7}'
)

MISC_ARGS=(
   --attention-dropout 0.0
   --hidden-dropout 0.0
   --accumulate-allreduce-grads-in-fp32
   --attention-softmax-in-fp32
   --attention-backend flash
)

DSPARK_ARGS=(
   --dspark-enabled
   --dspark-block-size 7
   --dspark-num-draft-layers 5
   --dspark-target-layer-ids 1,9,17,25,33
   --dspark-markov-rank 256
   --dspark-markov-head-type vanilla
   --dspark-num-anchors 512
   --dspark-mask-token-id 151669
   --dspark-ce-loss-alpha 0.5
   --dspark-l1-loss-alpha 0.5
   --dspark-confidence-head-alpha 0.1
   --dspark-loss-decay-gamma 4.0
   --dspark-draft-loss-weight 1.0
   --dspark-intermediate-size 9728
   --dspark-pretrained-model /data/nfs_87/xky/models/Qwen3-4B-dspark-pretrained/model.safetensors
)

# Launch ray
export MASTER_ADDR=${MASTER_ADDR:-"127.0.0.1"}
ray start --head --node-ip-address ${MASTER_ADDR} --num-gpus ${NUM_GPUS} --disable-usage-stats --dashboard-host=0.0.0.0 --dashboard-port=8265

VIME_DIR="/data/nfs_87/xky/dspark_new/vime"

RUNTIME_ENV_JSON="{
  \"working_dir\": \"${VIME_DIR}\",
  \"env_vars\": {
    \"PYTHONPATH\": \"/root/Megatron-LM/\",
    \"CUDA_DEVICE_MAX_CONNECTIONS\": \"1\",
    \"NCCL_NVLS_ENABLE\": \"${HAS_NVLINK}\"
  }
}"

ray job submit --address="http://127.0.0.1:8265" \
   --runtime-env-json="${RUNTIME_ENV_JSON}" \
   -- python3 train.py \
   --actor-num-nodes 1 \
   --actor-num-gpus-per-node ${NUM_GPUS} \
   --colocate \
   ${MODEL_ARGS[@]} \
   ${CKPT_ARGS[@]} \
   ${ROLLOUT_ARGS[@]} \
   ${OPTIMIZER_ARGS[@]} \
   ${GRPO_ARGS[@]} \
   ${PERF_ARGS[@]} \
   ${EVAL_ARGS[@]} \
   ${VLLM_ARGS[@]} \
   ${MISC_ARGS[@]} \
   ${DSPARK_ARGS[@]}

```


image

image

aoshen02:
Thanks for tracking this down. I think the current patch fixes the symptom, but the synchronization is placed too late and on a very hot path.

The root problem described here is that `torch_memory_saver.resume()` breaks the alias between each DSpark `param.data` and its Megatron DDP contiguous parameter buffer. Copying values after every `optimizer.step()` compensates for that broken alias indefinitely:

- it scans every chunk, buffer, and parameter on every training step;
- it puts a DSpark/TMS lifecycle workaround in the generic `forward_step()`;
- it depends on private/version-sensitive Megatron structures (`param_index_map`, `buckets`, and `param_to_index`);
- it may silently skip synchronization, and direct `ParamIndexSpec` unpacking is not valid in every Megatron version.

A more fundamental and smaller fix would be to restore the view once, immediately after TMS resume:

```python
torch_memory_saver.resume()

if self.args.dspark_enabled:
    restore_dspark_param_views(self.model)
```

The helper should live with the DSpark model integration (for example `dspark/modeling.py`), and rebind each draft parameter to its slice in the DDP contiguous parameter buffer. Once the alias is restored, the normal optimizer path updates draft parameters without an extra per-step copy.

Conceptually:

```python
def restore_dspark_param_views(model):
    for chunk in model:
        draft_params = set(unwrap_model(chunk).draft_model.parameters())
        for buffer in chunk.buffers:
            for param, spec in buffer.param_index_map.items():
                if param in draft_params:
                    param.data = buffer.param_data[
                        spec.start : spec.end
                    ].view_as(param)
```

The exact `ParamIndexSpec` fields should be taken from the pinned Megatron version rather than adding multi-version `hasattr` fallbacks.

The normalization correction looks right. Using `getattr(args, "dspark_enabled", False)` at the model-provider boundary is also reasonable because the provider is reused by Megatron conversion tooling with a different argument namespace.

I would also add a focused regression test instead of relying only on the 100-step convergence run:

1. capture the draft parameter and DDP-buffer relationship;
2. execute TMS pause/resume;
3. restore the views;
4. assert the draft parameter points at the expected buffer storage;
5. run one optimizer step and assert the draft parameter changes without any post-step copy.

That directly tests the reported failure mode and would make the fix both smaller and more first-principled.

PR #422, opened 2026-09-11
feat(ascend): support full and delta weight sync over disk

wangx700:
## Summary

Ascend currently selects only the colocated tensor updater or distributed collective updater. This adds non-colocated `--update-weight-transport disk` with both `--update-weight-mode full` and `--update-weight-mode delta`, based on ascend at 083aea63.

- Full mode gathers HF weights using Ascend's existing conversion iterator, writes safetensors shards and an index, then pauses rollout, flushes cache, reloads the published checkpoint and resumes generation.
- Delta mode captures the original HF checkpoint as version zero and publishes Zstandard-compressed XOR or overwrite deltas with resulting-tensor checksums. `VLLMEngine.pull_weights` uses `/collective_rpc` to materialize the checkpoint on every NPU worker host before disk reload.
- Port the local-checkpoint receiver and its tests from main's `docker/patch/latest/vllm-pull_weights.patch` into the existing NPU patch series, and add the NPU worker RPC entry point. Preserve the original model path as the baseline across reloads. Existing Dockerfile/series entries already apply both updated patches.
- Add `run-qwen3-4B-full-disk.sh` and `run-qwen3-4B-delta-disk.sh`, shared Qwen3-4B settings, configurable paths/Ray ports, and usage/flow documentation. These launchers require explicit device allocation and do not kill existing processes.
- Cap requested pinned pool storage at 8 GiB, release partial allocations on failure, and return buffers on worker-copy and pre-submission failures. Validate checkpoint headers, tensor offsets and short reads with file/tensor context. Bound receiver extraction to its own file diff in tests.

This adapts the disk protocol to Ascend's current updater lifecycle rather than importing main's newer RayTrainGroup architecture. The default collective path is unchanged; sparse HCCL and profiling changes are outside this PR.

## Validation

- Remote Ascend container, CPU tests: `python3 -m pytest -q tests/unit/backends/megatron_utils/update_weight/test_disk_weight_sync.py tests/unit/utils/test_disk_delta.py` — **29 passed**. Covers full/delta round trips, unchanged versions and repeated pulls, copy/submission failure injection, pinned allocation limits and cleanup, patch extraction boundaries, and missing/corrupted/truncated checkpoints. Pinned allocation behavior is simulated in CPU tests; this does not measure physical NPU host memory use.


vime/utils/arguments.py
```diff
@@ -1684,6 +1749,26 @@ def vime_validate_args(args):
     if args.save_interval is not None:
         assert args.save is not None, "'--save' is required when save_interval is set."
 
+    if args.update_weight_mode == "delta":
```
momo609, line 1752: Are these added parameters aligned with the main architecture? If they are extra, what prevents them from being reused?

